PR #25, opened 2017-05-30
Changed naming for some pages, no longer stores files by name on server

abhinadduri:


app.js
```diff
@@ -26,12 +26,12 @@ app.get('/assets/download/:id', function(req, res) {
...
   if (!reply) {
-      res.send('error');
+      res.send('This link has expired!');
     } else {
       res.setHeader('Content-Disposition', 'attachment; filename=' + reply);
       res.setHeader('Content-Type', 'application/octet-stream');
-
-      res.download(__dirname + '/static/' + reply);
+      client.del(id);
+      res.download(__dirname + '/static/' + id, reply);
```
dannycoates: Let's move the `client.del(id)` into the callback of `download`. We'll get an error there if something went wrong, and at that point we can also delete the file off the disk.

Something like:
```js
res.download(__dirname + '/static/' + id, reply, function (err) {
  if (!err) {
    client.del(id)
    fs.unlinkSync(__dirname + '/static/' + id)
  }
  // log the error
})
```

app.js
```diff
@@ -26,12 +26,12 @@ app.get('/assets/download/:id', function(req, res) {
   let id = req.params.id;
   client.hget(id, "filename", function(err, reply) { // maybe some expiration logic too
   if (!reply) {
-      res.send('error');
+      res.send('This link has expired!');
```
dannycoates: We should return a 404 in this case.

PR #28, opened 2017-05-31
Added a UI for the uploader end, made stylistic changes, implemented deleting

abhinadduri:


app.js
```diff
@@ -1,72 +1,98 @@
...
+                let uuid = Math.floor(Math.random() * 10000000).toString()
+                         + Math.floor(Math.random() * 10000000).toString()
+                         + Math.floor(Math.random() * 10000000).toString();
+
+                client.hset(id, "filename", filename);
+                client.hset(id, "expiration", 0);
+                client.hset(id, "delete", uuid);
+
+                // delete the file off the server in 24 hours
+                setTimeout(function() {
```
dannycoates: It's not ideal to have a bunch of timeout callbacks hanging around like this. We'll need to find another way to do this before launch. Maybe we can rely on expiring them in S3, otherwise a separate process from the server to do the cleanup would be ok.

public/upload.js
```diff
@@ -0,0 +1,150 @@
...
+function strToIv(str) {
+  var iv = new Uint8Array(16);
+  for (var i = 0; i < str.length; i += 2) {
+    iv[i/2] = parseInt((str.charAt(i) + str.charAt(i + 1)), 16);
+  }
+
+  return iv;
+}
+
+function returnBindedLI(a_element, name, link, li) {
```
dannycoates: The ui code here is fine for now, but eventually we'll want to refactor both upload and download to return objects that emit their own events so we can decouple them from the ui.

ericawright:
nit: I noticed all the changes to double quotes  -- I assume this is because of the proposed eslint rule
 `quotes: [error, single]`
this actually means that in javascript we want to only use single quotes.

dannycoates:
@abhinadduri let's add eslint asap after this

PR #37, opened 2017-06-01
Refactoring

abhinadduri:


package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "format": "prettier --single-quote --write 'public/*.js' 'app.js'",
+    "format": "prettier --single-quote --write 'frontend/src/*.js'",
```
pdehaan: You'll also need to include the new 'server/' directory here:

```
"format": "prettier --single-quote --write 'frontend/src/*.js' 'server/*.js'",
```

package.json
```diff
@@ -17,5 +17,9 @@
     "path": "^0.12.7",
     "prettier": "^1.3.1",
     "redis": "^2.7.1"
+  },
+  "devDependencies": {
+    "browserify": "^14.4.0",
+    "watchify": "^3.9.0"
```
pdehaan: Not related to anything, but we can probably move **prettier** to a devDep.

public/index.html
```diff
@@ -1,8 +1,8 @@
 <!DOCTYPE html>
 <html>
 <head>
-<title>Firefox Fileshare</title>
-<script src="upload.js"></script>
+  <title>Firefox Fileshare</title>
+  <script src="bundle.js"></script>
```
pdehaan: Not sure if we want this as "bundle.js" or "/bundle.js" (to be consistent w/ _public/download.html_ above)...

frontend/src/upload.js
```diff
@@ -0,0 +1,156 @@
+const EventEmitter = require('events');
```
pdehaan: 'EventEmitter' is assigned a value but never used

frontend/src/upload.js
```diff
@@ -0,0 +1,156 @@
...
+      hexStr += '0' + iv[i].toString(16);
+    } else {
+      hexStr += iv[i].toString(16);
+    }
+  }
+  window.hexStr = hexStr;
+  return hexStr;
+};
+
+let strToIv = str => {
```
pdehaan: I think this is unused here.
It looks like it may also be defined in frontend/src/download.js:114-121.

frontend/src/download.js
```diff
@@ -0,0 +1,142 @@
...
+
+    if (this.status == 200) {
+      let self = this;
+      let blob = new Blob([this.response]);
+      let arrayBuffer;
+      let fileReader = new FileReader();
+      fileReader.onload = function() {
+        arrayBuffer = this.result;
+        let array = new Uint8Array(arrayBuffer);
+        salt = strToIv(location.pathname.slice(10, -1));
```
pdehaan: I think we need to declare `salt` somewhere with `let` or `const`.

frontend/src/download.js
```diff
@@ -0,0 +1,142 @@
...
+  xhr.addEventListener('progress', updateProgress.bind(null, listelem));
+
+  xhr.onload = function(e) {
+    // maybe send a separate request before this one to get the filename?
+
+    // maybe render the html itself with the filename, since it's generated server side
+    // after a get request with the unique id
+    listelem.emit(
+      'name',
+      xhr.getResponseHeader('Content-Disposition').match(/filename="(.+)"/)[1]
```
pdehaan: Weird question, but would it ever be possible that `.match()` wouldn't find a match and return `null` and then implode on accessing `[1]`? Do we need a `try..catch` or some error handling?

dannycoates: Possibly. We've got a lot of error handling left to do, this included. Right now we're only thinking happy thoughts :)

PR #42, opened 2017-06-02
changed to handle 404 during download, also removing progress listene…

abhinadduri:
…r once percent completion is 100 to avoid manipulating DOM twice

### DIFF
--- a/frontend/src/fileReceiver.js
+++ b/frontend/src/fileReceiver.js
@@ -20,6 +20,11 @@ class FileReceiver extends EventEmitter {
         };
 
         xhr.onload = function(e) {
+
+          if (xhr.status === 404) {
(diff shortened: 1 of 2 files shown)


frontend/src/fileReceiver.js
```diff
@@ -20,6 +20,11 @@ class FileReceiver extends EventEmitter {
         };
 
         xhr.onload = function(e) {
+
+          if (xhr.status === 404) {
+            reject(new Error('The file has expired, or has already been deleted.'));
```
pdehaan: Does this need a `return` statement to prevent leaking into the code path(s) below?

dannycoates: yes, good catch @pdehaan

abhinadduri: added the return

PR #45, opened 2017-06-06
S3 integration

abhinadduri:


dannycoates:
We should keep the file system version so we can dev without aws credentials. Lets factor out the storage bits into its own module. I sketched something up as a start here: https://gist.github.com/dannycoates/f75e37df25968ff74475156b894669d4

server/storage.js
```diff
@@ -0,0 +1,208 @@
...
+  module.exports = {
+    length: LocalLength,
+    get: LocalGet,
+    set: LocalSet,
+    delete: LocalDelete,
+    forceDelete: LocalForceDelete
+  };
+}
+
+function LocalLength(id) {
```
dannycoates: Traditionally, in js this case style is used for constructors. Let's use camelCase for these function names

server/portal_server.js
```diff
@@ -4,12 +4,24 @@ const path = require('path');
...
+  conf.bitly_key !== 'localhost';
+
+const AWS = require('aws-sdk');
+const s3 = new AWS.S3();
 
 const app = express();
-const redis = require('redis'),
-  client = redis.createClient();
+const redis = require('redis');
+const redis_client = redis.createClient();
```
dannycoates: can we move all this redis code into `storage.js`?

server/portal_server.js
```diff
@@ -4,12 +4,24 @@ const path = require('path');
...
+const fetch = require('node-fetch');
+const storage = require('./storage.js');
+
+let isProduction =
+  conf.env === 'production' &&
+  conf.s3_bucket !== 'localhost' &&
+  conf.bitly_key !== 'localhost';
+
+const AWS = require('aws-sdk');
+const s3 = new AWS.S3();
```
dannycoates: We can remove `AWS` and `s3` from this file now

server/portal_server.js
```diff
@@ -4,12 +4,24 @@ const path = require('path');
 const fs = require('fs-extra');
 const bodyParser = require('body-parser');
 const crypto = require('crypto');
+const conf = require('./config.js');
+const stream = require('stream');
+const fetch = require('node-fetch');
+const storage = require('./storage.js');
+
+let isProduction =
```
dannycoates: It's confusing that `isProduction` can be `false` when `NODE_ENV=production`. Perhaps naming it something like `useRemoteServices` or `localhostOnly`.

server/storage.js
```diff
@@ -0,0 +1,208 @@
...
+      });
+    });
+  });
+}
+
+function LocalDelete(id, client, delete_token) {
+  return new Promise((resolve, reject) => {
+    client.hget(id, 'delete', (err, reply) => {
+      if (!reply || delete_token !== reply) {
+        resolve(
```
dannycoates: this block can just be `reject()`

server/storage.js
```diff
@@ -0,0 +1,208 @@
...
+  let params = {
+    Bucket: conf.s3_bucket,
+    Key: id,
+    Body: file
+  };
+
+  return new Promise((resolve, reject) => {
+    s3.upload(params, function(err, data) {
+      if (err) {
+        console.log(err, err.stack); // an error occurred
```
dannycoates: We should call `reject()` here

server/storage.js
```diff
@@ -0,0 +1,208 @@
...
+    client.hget(id, 'delete', (err, reply) => {
+      if (!reply || delete_token !== reply) {
+        resolve(
+          new Promise((resolve, reject) => {
+            reject();
+          })
+        );
+      } else {
+        resolve(
+          new Promise((resolve, reject) => {
```
dannycoates: we shouldn't need this `Promise`, just resolve the unlink

server/portal_server.js
pdehaan: Nit: I don't think this variable is used in this file anymore.
But @dannycoates comment still applies to server/storage.js:10 below.

dannycoates:
This looks good to merge. I don't think we necessarily need index to be a template but NBD

Issue #27, opened 2017-05-31
Lint


ericawright:
- Add a linter so that we all follow the same coding conventions
https://github.com/eslint/eslint
https://github.com/stylelint/stylelint

- Protect master branch, and hook up Travis or Circle CI to check that it passes for PRs

pdehaan:
For ESLint, probably something like this (at a minimum):

```yaml
env:
  browser: true
  es6: true
  node: true

extends:
  - eslint:recommended

rules:
  eqeqeq: error
  no-console: warn
  no-var: error
  one-var: [error, never]
  prefer-const: error
  quotes: [error, single]
```

pdehaan:
I haven't used **stylelint** a bunch, but this seems to work as a minimal config (plus, latest version of **prettier** also formats CSS files):

```diff
  "devDependencies": {
    ...
    "prettier": "1.4.4",
+   "stylelint": "7.10.1",
+   "stylelint-config-standard": "16.0.0",
    ...
  },
  "scripts": {
+   "format:css": "prettier --write 'public/*.css'",
+   "lint:css": "stylelint 'public/*.css' --fix",
    ...
  }
```

And then a .stylelintrc config file of:

```yaml
extends:
  - stylelint-config-standard

rules:
  selector-list-comma-newline-after: null
```

pdehaan:
Looking at this a bit closer, we _may_ want to add some basic HTML linting as well:

```sh
➜  something-awesome git:(master) $ npm run lint:html

> portal-alpha@1.0.0 lint:html /Users/pdehaan/dev/github/mozilla/something-awesome
> htmllint 'views/*.handlebars'

views/download.handlebars: line 19, col 7, tag is not closed
views/download.handlebars: line 23, col 9, the `alt` property must be set for image tags
views/download.handlebars: line 34, col 9, the `alt` property must be set for image tags
views/index.handlebars: line 18, col 30, the `alt` property must be set for image tags
views/index.handlebars: line 39, col 12, the `width` attribute is banned
views/index.handlebars: line 39, col 18, the `width` attribute is not double quoted
views/index.handlebars: line 40, col 12, the `width` attribute is banned
views/index.handlebars: line 40, col 18, the `width` attribute is not double quoted
views/index.handlebars: line 41, col 12, the `width` attribute is banned
views/index.handlebars: line 41, col 18, the `width` attribute is not double quoted
views/index.handlebars: line 42, col 12, the `width` attribute is banned
[...]

PR #54, opened 2017-06-07
Adding basic ESLint config

pdehaan:
This only adds ESLint, not CSS/style linting.

Ref: #27

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+public/bundle.js
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,38 @@
+env:
+  browser: true
+  es6: true
+  jquery: true
+  node: true
+
+extends:
+  - eslint:recommended
+  - plugin:node/recommended
+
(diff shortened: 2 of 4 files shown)


ericawright:
Looks good. But prettier is already in, we don't need both, right? I haven't figured out how to customize prettier yet, so I prefer this ;)

pdehaan:
Well... I think **prettier** is more for code _formatting_, whereas **eslint** is better suited for code _validation_ and enforcing rules. Definitely some areas of overlap (indentation, quotes, etc) though.

We _could_ use **prettier** as a plugin for **eslint** using something like [**eslint-plugin-prettier**](https://github.com/prettier/eslint-plugin-prettier), but I find that flow a bit awkward and difficult to maintain.

dannycoates:
Thanks @pdehaan!

@abhinadduri @dnarcese let's get this merged after #71 then create an issue to fix the lint errors

pdehaan:
@dannycoates Rebased. 👀

PR #71, opened 2017-06-08
Drag & drop

### LINKED ISSUE #61: Implement drag & drop

dnarcese:
Added drag & drop as well as a delete file confirmation.

Closes #61

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -42,7 +42,7 @@ $(document).ready(function() {
     } else {
       file = event.target.files[0];
     }
-    let fileList = $('#uploaded-files');
+    let $fileList = $('#uploaded-files');
     let row = document.createElement('tr');
     let name = document.createElement('td');
     let link = document.createElement('td');
@@ -74,7 +74,7 @@ $(document).ready(function() {
     $(popupDiv).append($popupText);
     del.appendChild(popupDiv);
     row.appendChild(del);
-    fileList.append(row); //add row to table
+    $fileList.append(row); //add row to table
 
     const fileSender = new FileSender(file);
     fileSender.on('progress', percentComplete => {
@@ -91,28 +91,32 @@ $(document).ready(function() {
       localStorage.setItem(info.fileId, info.deleteToken);
 
       // delete file
-      console.log($popupText.first());
       $popupText.find('.del-file').click(e => {
         FileSender.delete(
           info.fileId,
           localStorage.getItem(info.fileId)
-        ).then(() => {
-          e.target.parentNode.parentNode.parentNode.parentNode.remove();
+       ).then(() => {
+          //
+          $(e.target).parents('tr').remove();
           localStorage.removeItem(info.fileId);
         });
       });
 
       // show popup
       btn.addEventListener('click', e => {
-        $popupText.toggleClass('show');
+        toggleShow();
       });
       // hide popup
       $popupText.find('.nvm').click(e => {
-        $popupText.toggleClass('show');
+        toggleShow();
       });
       $('#upload-progress').hide();
       $('#share-link').show();
     });
+    
+    function toggleShow(){
+      $popupText.toggleClass('show');
+    }
   };
 
   window.allowDrop = function(ev) {
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -36,8 +36,8 @@ $(document).ready(function() {
   // on file upload by browse or drag & drop
   window.onUpload = event => {
     event.preventDefault();
-    let file = "";
-    if (event.type == "drop"){
+    let file = '';
+    if (event.type == 'drop') {
       file = event.dataTransfer.files[0];
     } else {
       file = event.target.files[0];
@@ -50,7 +50,7 @@ $(document).ready(function() {
     let del = document.createElement('td');
     let btn = document.createElement('button');
     let popupDiv = document.createElement('div');
-    let $popupText = $('<span>', {'class': 'popuptext'});
+    let $popupText = $('<span>', { 'class': 'popuptext' });
     let cellText = document.createTextNode(file.name);
     let progress = document.createElement('p');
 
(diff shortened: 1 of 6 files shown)


frontend/src/upload.js
```diff
@@ -62,25 +89,33 @@ $(document).ready(function() {
...
       link.innerHTML = url;
       localStorage.setItem(info.fileId, info.deleteToken);
-      let del = document.createElement('td');
-      let btn = document.createElement('button');
-      btn.innerHTML = 'x';
-      btn.classList.add('delete-btn');
-      btn.addEventListener('click', e => {
+
+      // delete file
+      console.log($popupText.first());
```
ericawright: nit: remove the console.logs

frontend/src/upload.js
```diff
@@ -62,25 +89,33 @@ $(document).ready(function() {
...
+      $popupText.find('.del-file').click(e => {
         FileSender.delete(
           info.fileId,
           localStorage.getItem(info.fileId)
         ).then(() => {
-          e.target.parentNode.parentNode.remove();
+          e.target.parentNode.parentNode.parentNode.parentNode.remove();
```
ericawright: With jQuery you could try `$(e.target).parents('tr')`
with vanilla JS `e.target.closest('tr')`

frontend/src/upload.js
```diff
@@ -62,25 +89,33 @@ $(document).ready(function() {
...
-      row.appendChild(del);
+
+      // show popup
+      btn.addEventListener('click', e => {
+        $popupText.toggleClass('show');
+      });
+      // hide popup
+      $popupText.find('.nvm').click(e => {
+        $popupText.toggleClass('show');
+      });
```
ericawright: these two have the same fn after the click event, perhaps we name the fn and reuse it?

views/index.handlebars
```diff
@@ -41,7 +41,9 @@
           <th width=18%>Expires in</th>
           <th width=7%>Delete</th>
         </tr>
-        <div data-role="popup" id="popupArrow" data-arrow="true">
+        <!-- <div class="popup">
```
ericawright: nit: remove commented code

frontend/src/upload.js
```diff
@@ -29,25 +33,48 @@ $(document).ready(function() {
...
+  // on file upload by browse or drag & drop
+  window.onUpload = event => {
+    event.preventDefault();
+    let file = '';
+    if (event.type == 'drop') {
+      file = event.dataTransfer.files[0];
+    } else {
+      file = event.target.files[0];
+    }
     let fileList = $('#uploaded-files');
```
ericawright: nit `$` in front to follow jquery convention

public/main.css
```diff
@@ -108,6 +108,59 @@ td {
...
+   background-color: #555;
+   color: #fff;
+   text-align: center;
+   border-radius: 6px;
+   padding: 8px 0;
+   position: absolute;
+   z-index: 1;
+   bottom: 125%;
+   left: 50%;
+   margin-left: -80px;
```
ericawright: do you need both left and margin-left here?

public/main.css
```diff
@@ -108,6 +108,59 @@ td {
...
+   margin-left: -5px;
+   border-width: 5px;
+   border-style: solid;
+   border-color: #555 transparent transparent transparent;
+}
+
+/* Toggle this class when clicking on the popup container (hide and show the popup) */
+.popup .show {
+   visibility: visible;
+   -webkit-animation: fadeIn 1s;
```
ericawright: shouldn't need `-webkit`, it's old syntax
Can we change this to a transition, my understanding is that it is more efficient than a keyframes animation.

frontend/src/upload.js
```diff
@@ -29,25 +33,48 @@ $(document).ready(function() {
...
 
-  let onChange = event => {
-    const file = event.target.files[0];
-
+  // on file upload by browse or drag & drop
+  window.onUpload = event => {
+    event.preventDefault();
+    let file = '';
+    if (event.type == 'drop') {
+      file = event.dataTransfer.files[0];
```
pdehaan: Does this work if I drop multiple files on the drop target from Finder? Or will it only handle the first image?

server/portal_server.js
```diff
@@ -100,6 +100,7 @@ app.post('/delete/:id', (req, res) => {
     .then(err => {
       if (!err) {
         console.log('Deleted.');
+        res.sendStatus(200);
```
dannycoates: nice catch :)

frontend/src/upload.js
dannycoates: Good question. If it doesn't we can follow up with a solution in another PR

pdehaan: Spoiler alert: It doesn't.

But it looks like we can do `Array.from().forEach()` (or just a regular `for` loop) on the FileList as a workaround. But that exposes another dull edge case where you're dragging 2-3 files onto the app and it returns you a single share URL that you can copy to your clipboard.

```js
  window.onUpload = event => {
    event.preventDefault();
    let fileList = $('#uploaded-files');
    let files = [];

    if (event.type === 'drop') {
      files = event.dataTransfer.files;
    } else {
      files = event.target.files;
    }

    Array.from(files).forEach(file => { ... });
```

frontend/src/upload.js
```diff
@@ -63,25 +90,37 @@ $(document).ready(function() {
...
       });
-      del.appendChild(btn);
-      row.appendChild(del);
+
+      // show popup
+      btn.addEventListener('click', e => {
+        toggleShow();
+      });
+      // hide popup
+      $popupText.find('.nvm').click(e => {
```
ericawright: btn.addEventListener('click', toggleShow);

      $popupText.find('.nvm').click(toggleShow);

ericawright:
just the one more nit and then I think we're good to merge :)

PR #72, opened 2017-06-08
Logging

abhinadduri:
Added logging with mozlog, changed jquery from ajax script tag, added exists route

### DIFF
--- a/server/log.js
+++ b/server/log.js
@@ -1,0 +1,15 @@
+const conf = require('./config.js');
+
+let notLocalHost =
+  conf.env === 'production' &&
+  conf.s3_bucket !== 'localhost' &&
+  conf.bitly_key !== 'localhost';
+
+const mozlog = require('mozlog') ({
+  app: 'FirefoxFileshare',
+  level: notLocalHost ? 'INFO' : 'verbose',
+  fmt: notLocalHost ? 'heka' : 'pretty',
+  debug: !notLocalHost
+})
+
+module.exports = mozlog;
--- a/server/log.js
+++ b/server/log.js
@@ -1,9 +1,6 @@
 const conf = require('./config.js');
 
-let notLocalHost =
-  conf.env === 'production' &&
-  conf.s3_bucket !== 'localhost' &&
-  conf.bitly_key !== 'localhost';
+let notLocalHost = conf.notLocalHost;
 
 const mozlog = require('mozlog') ({
   app: 'FirefoxFileshare',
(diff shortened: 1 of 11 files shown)


server/log.js
```diff
@@ -0,0 +1,15 @@
+const conf = require('./config.js');
+
+let notLocalHost =
+  conf.env === 'production' &&
+  conf.s3_bucket !== 'localhost' &&
+  conf.bitly_key !== 'localhost';
```
pdehaan: This snippet seems to be in 3-4 places. May want to refactor it out into a utils file or something.

server/portal_server.js
```diff
@@ -113,8 +125,9 @@ app.post('/upload/:id', (req, res, next) => {
 
   req.pipe(req.busboy);
   req.busboy.on('file', (fieldname, file, filename) => {
-    console.log('Uploading: ' + filename);
-    const protocol = conf.env === 'development' ? req.protocol : 'https';
+    log.info('Uploading:', req.params.id);
+
+    const protocol = !notLocalHost ? req.protocol : 'https';
```
dannycoates: how about `const protocol = notLocalHost ? 'https' : req.protocol;` to eliminate the double-negative?

server/portal_server.js
```diff
@@ -63,18 +74,19 @@ app.get('/assets/download/:id', (req, res) => {
...
-        let file_stream = storage.get(id);
-
-        file_stream.on(notLocalHost ? 'finish' : 'close', () => {
-          storage.forceDelete(id).then(err => {
-            if (!err) {
-              console.log('Deleted.');
-            }
-          });
-        });
+      });
```
dannycoates: the stream pipe needs to happen after the writeHead otherwise the pipe can cause the default headers to get sent and cause writeHead to throw an error.

server/storage.js
```diff
@@ -53,6 +63,14 @@ function filename(id) {
   });
 }
 
+function localExists(id) {
+  return new Promise((resolve, reject) => {
+    redis_client.hget(id, 'filename', (rediserr, reply) => {
+      resolve(fs.existsSync(__dirname + '/../static/' + id) && !rediserr)
```
dannycoates: It's enough to just check if the id is in redis. same with s3

dannycoates: Like `redis_client.exists(id ...`

server/log.js
dannycoates: how about we put this in `config.js` and attach it to conf, so you'd use it like `conf.notLocalHost`?

abhinadduri: sure!

server/portal_server.js
abhinadduri: good idea

PR #79, opened 2017-06-09
Optimize/minimize bundle.js for production

### LINKED ISSUE #78: Optimize/minimize bundle.js
I randomly noticed that the current bundle has an inline `//# sourceMappingURL=data:application/json;charset=utf-8;base64,` at the bottom of the file.

Not bad for development, but I did notice that it makes the file explode up to 744KB. The same file _without_ the inline source map is about 281KB, which is a lot more mobile friendly.

pdehaan:
Fixes #78

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -533,8 +533,7 @@
     "commander": {
       "version": "2.9.0",
       "resolved": "https://registry.npmjs.org/commander/-/commander-2.9.0.tgz",
-      "integrity": "sha1-nJkJQXbhIkDLItbFFGCYQA/g99Q=",
-      "dev": true
+      "integrity": "sha1-nJkJQXbhIkDLItbFFGCYQA/g99Q="
     },
     "concat-map": {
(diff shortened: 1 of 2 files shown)


pdehaan:
### `$ npm start` _(146 KB)_:
```sh
$ ls -lash public/*.js
146K Jun  9 13:33 public/bundle.js
 22K Jun  9 08:33 public/webcrypto-shim.js
```

### `npm run dev` _(744 KB -- w/ inline sourcemap)_:
```sh
$ ls -lash public/*.js
774K Jun  9 13:38 public/bundle.js
 22K Jun  9 08:33 public/webcrypto-shim.js
```

pdehaan:
Disclaimer: This junk may or may not work. Browserify has never played nice w/ Uglify. That was supposed to be fixed via [**uglify-es**](http://npm.im/uglify-es), but I think npm5 is giving me semi-inconsistent results depending on how things get written to the package-lock.json.

🤷‍♀️

PR #93, opened 2017-06-13
Bump ESLint and eslint-plugin-security deps

pdehaan:


.eslintrc.yml
```diff
@@ -15,18 +16,8 @@ plugins:
...
-  security/detect-disable-mustache-escape: warn
-  security/detect-eval-with-expression: warn
-  security/detect-new-buffer: warn
-  security/detect-no-csrf-before-method-override: warn
-  security/detect-non-literal-fs-filename: warn
-  security/detect-non-literal-regexp: warn
-  security/detect-non-literal-require: warn
-  security/detect-possible-timing-attacks: warn
-  security/detect-pseudoRandomBytes: warn
-  security/detect-unsafe-regex: warn
```
pdehaan: All these default rules are enabled in _"plugin:security/recommended"_ now (as of v1.4.0).

.eslintrc.yml
```diff
@@ -15,18 +16,8 @@ plugins:
...
+  security/detect-non-literal-fs-filename: off
+  security/detect-object-injection: off
```
pdehaan: Disable these 2 rules since the non-literal filesystem filenames are only for local development, and the other rule is kinda noisy.

pdehaan:
Closing in favor of more aggressive #104

PR #99, opened 2017-06-19
Testing

abhinadduri:
Finished testing for storage modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,7 @@
     "mozlog": "^2.1.1",
     "node-fetch": "^1.7.1",
     "path": "^0.12.7",
-    "proxyquire": "^1.8.0",
-    "redis": "^2.7.1",
-    "sinon": "^2.3.2"
+    "redis": "^2.7.1"
   },
   "devDependencies": {
     "browserify": "^14.4.0",
@@ -29,11 +27,13 @@
     "eslint-plugin-node": "5.0.0",
     "eslint-plugin-security": "1.3.0",
     "htmllint-cli": "0.0.6",
+    "mocha": "^3.4.2",
     "npm-run-all": "4.0.2",
     "prettier": "1.4.4",
+    "proxyquire": "^1.8.0",
+    "sinon": "^2.3.5",
     "stylelint": "7.11.0",
     "stylelint-config-standard": "16.0.0",
-    "mocha": "^3.4.2",
     "watchify": "^3.9.0"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "mozlog": "^2.1.1",
     "node-fetch": "^1.7.1",
     "path": "^0.12.7",
-    "redis": "^2.7.1"
+    "proxyquire": "^1.8.0",
+    "redis": "^2.7.1",
+    "sinon": "^2.3.2"
   },
   "devDependencies": {
     "browserify": "^14.4.0",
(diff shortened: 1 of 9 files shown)


test/aws.storage.test.js
```diff
@@ -0,0 +1,140 @@
...
+    assert(storage.get('123') === null);
+  })
+});
+
+describe('Testing Set using aws', function() {
+  it('Should pass when the file is successfully uploaded and no bitly key', function() {
+    conf.bitly_key = null;
+    s3Stub.upload.callsArgWith(1, null, {});
+    return storage.set('123', {}, 'Filename.moz', 'url.com')
+           .then(reply => assert(reply.url === 'url.com' && reply.uuid !== null))
```
dannycoates: Since this function has multiple side effects it's useful to assert that the ones we expect to get called actually were. For example, setting the expire time in redis. We can use sinon to assert that it was called, and with the arguments we expect:
```js
assert(expire.calledOnce)
assert(expire.calledWith('123', 86400000))
```
ref: http://sinonjs.org/releases/v2.3.4/spies/

For this to work you might need to create a [beforeEach](http://mochajs.org/#hooks) to call `expire.reset()` etc.

test/aws.storage.test.js
```diff
@@ -0,0 +1,140 @@
...
+  'fs': fsStub,
+  './log.js': function() { return logStub },
+  'aws-sdk': awsStub
+});
+
+describe('Testing Length using aws', function() {
+  it('Filesize returns properly if id exists', function() {
+    s3Stub.headObject.callsArgWith(1, null, {ContentLength: 1});
+    return storage.length('123')
+                  .then(reply => assert(reply === 1))
```
dannycoates: lets prefer `assert.equal` when possible. It'll print better messages when the assertion fails, including the actual and expected values.

test/aws.storage.test.js
```diff
@@ -0,0 +1,140 @@
...
+  it('Should not error out when the file exists', function() {
+    s3Stub.getObject.returns({
+      createReadStream: function() { return 1; }
+    });
+    assert(storage.get('123') === 1);
+  })
+
+  it('Should error when the file does not exist', function() {
+    s3Stub.getObject.returns({
+      createReadStream: function() { return null; }
```
dannycoates: this can be a spy that throws to more closely match the actual behavior

pdehaan:
We should also tweak the .eslintrc.yml file to recognize the **mocha** specific keywords:

```diff
diff --git a/.eslintrc.yml b/.eslintrc.yml
index 16f7f63..b98f50c 100644
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -2,6 +2,7 @@ env:
   browser: true
   es6: true
   jquery: true
+  mocha: true
   node: true

 extends:
```

server/portal_server.js
```diff
@@ -101,7 +103,7 @@ app.post('/delete/:id', (req, res) => {
 
   storage
     .delete(id, delete_token)
-    .then(err => {
+    .then(() => {
       if (!err) {
```
pdehaan: I think `err` is now undefined here.

package.json
```diff
@@ -17,7 +17,9 @@
     "mozlog": "^2.1.1",
     "node-fetch": "^1.7.1",
     "path": "^0.12.7",
-    "redis": "^2.7.1"
+    "proxyquire": "^1.8.0",
+    "redis": "^2.7.1",
+    "sinon": "^2.3.2"
```
abhinadduri: moved them to dev dependencies

test/aws.storage.test.js
abhinadduri: refactored

Issue #86, opened 2017-06-13
Update encryption format to use gcm


wresuolc:
We should be using GCM.  Details are at  https://www.w3.org/TR/WebCryptoAPI/#aes-gcm .  Use a 256 bits key + additional data (aad) + IV, all random and unique to each file.  Put the aad and iv in the file metadata, keep the key secret in the fragment.

Ping @jvehent for review

dannycoates:
GCM means we'll work on fewer browsers, right? Is that a worthwhile trade off? I don't deeply understand what disadvantage we're trying to avoid from CBC. I know that it's "less recommended" than GCM but in this use case of 24h storage is it a deal breaker? I'd prefer to support as many browsers (and versions, including mobile) as reasonable.

jvehent:
The main benefit of GCM over CBC is the added layer of authentication that guarantees the integrity of the message. With CBC, you could decrypt something with the key, thinking you got the original, only to end up with data that's completely different from what you'd expect. GCM's authentication prevents that.

You could achieve the same authentication with CBC by adding an HMAC with the secret key, but I think using the current crypto standard instead of the old one sends a better message.

I don't know about browser compatibility. This is a valid point I cannot answer.

jvehent:
Another option might be to use a standard encryption format, like JWE: https://tools.ietf.org/html/rfc7516

jvehent:
See also https://github.com/mozilla/notes/pull/77 for an example of using the `js-jose` library to encrypt data with JWE AES256-GCM.

abhinadduri:
@jvehent I have pushed a 256 bit gcm encryption scheme to the gcm branch. Please check it out and let me know if there are changes that need to be made. I've also gotten it to work on all modern browsers (Chrome, Firefox, Edge, and Safari). I have not tested it on Opera yet.

jvehent:
@abhinadduri Could you use JWE with the `js-jose` library? It's an encryption standard that would default to aes256-gcm and be easier to integrate in other libraries.

dannycoates:
JWE (and js-jose in particular) isn't ideal for our use case of encrypting large files. I did up a test with it and the amount of string conversion and manipulation made it *very* slow. The one part it made nicer was that key parameters were handled for you.

[...]

PR #106, opened 2017-06-20
Gcm

abhinadduri:


pdehaan:
Fixes #86?

ekr:
Can you sketch out what's going on with the crypto, divorced from the code?

It looks like you're using a fraction of the pathname as a "salt" which you then call the IV? What properties do you expect that to have?

abhinadduri:
@ekr The iv is also used as the file id, which is what we use to retrieve the file in the server-side code. The salt is not a parameter sent to the server (it follows the hash), and is used to encrypt/decrypt files. The aad is a separate field sent to the server by the uploader, for the sake of additional security. 

All of these values are random Uint8's created by the window.crypto module. 

The uploader uses the iv, salt, and aad to encrypt a file. The server sends the encrypted file to a receiver, who uses the iv and salt from the copy/pasted url, and the aad it gets from the server to decrypt the file in the browser.

ekr:
@abhinadduri I'm not really following. Do you have a wire protocol specification I can look at?

server/portal_server.js
```diff
@@ -136,20 +137,24 @@ app.post('/delete/:id', (req, res) => {
 });
 
 app.post('/upload/:id', (req, res, next) => {
```
dannycoates: We no longer need the `:id` parameter in the upload url. `new_id` should be created at this level instead of in storage.js. It'd also be better to create `meta.delete` up here too. Ideally storage.js should only be concerned with saving and fetching the data from persistent storage and leaving the logic of what the data is and how it's created to something else (this function).

frontend/src/fileReceiver.js
```diff
@@ -1,12 +1,10 @@
...
-const { strToIv } = require('./utils');
-
-const Raven = window.Raven;
+const { hexToArray } = require('./utils');
 
 class FileReceiver extends EventEmitter {
   constructor() {
     super();
-    this.salt = strToIv(location.pathname.slice(10, -1));
+    this.salt = hexToArray(location.pathname.slice(10, -1));
```
dannycoates: `iv` is coming from `X-File-Metadata` now so we shouldn't need this anymore.

frontend/src/fileSender.js
```diff
@@ -37,43 +37,54 @@ class FileSender extends EventEmitter {
...
+              {
+                name: 'AES-GCM',
+                iv: this.iv,
+                additionalData: file.hash,
+                tagLength: 128
+              },
+              secretKey,
+              file.plaintext
+            )
+            .catch(err => console.log('Error with encrypting.')),
```
dannycoates: this needs to rethrow

server/storage.js
```diff
@@ -82,24 +116,20 @@ function localGet(id) {
...
+function localSet(id, newId, file, filename, meta) {
   return new Promise((resolve, reject) => {
-    const fstream = fs.createWriteStream(path.join(__dirname, '../static', id));
+    const fstream = fs.createWriteStream(path.join(__dirname, '../static', newId));
     file.pipe(fstream);
     fstream.on('close', () => {
-      const uuid = crypto.randomBytes(10).toString('hex');
-
-      redis_client.hmset([id, 'filename', filename, 'delete', uuid]);
+      meta.id = id;
```
dannycoates: what is `meta.id` being used for? We probably don't need the old `id` that used to be `iv` anymore. If so we can remove it as a function parameter too

PR #115, opened 2017-06-22
added error checking to front end code, set up sentry

abhinadduri:


public/raven.min.js
```diff
@@ -0,0 +1,3 @@
...
+//# sourceMappingURL=raven.min.js.map
```
pdehaan: @dannycoates Do we want to install [**raven**](https://www.npmjs.com/package/raven) via npm and use some script to copy the raven.min.js file into _/public/*_? Adding vendor code to repos always makes me uneasy.

frontend/src/fileSender.js
```diff
@@ -64,6 +65,7 @@ class FileSender extends EventEmitter {
           window.crypto.subtle.exportKey('jwk', secretKey)
         ]);
       })
+      .catch(err => { Raven.captureException(err) })
```
pdehaan: This _may_ have some side effects, IIRC.
I think `.catch()` will handle/swallow any errors and then proceed onto the next `.then()`, which may get some null/undefined args.
So some things may fail, but not 100% sure what will happen and if it may rethrow errors and trigger the `.catch()` on L105 below and resend some errors to Raven.

abhinadduri: fixed in 8c1ccb9

public/raven.min.js
abhinadduri: Alternatively we can include it as a cdn. Thoughts?

abhinadduri: Another way would be to use another route, i.e. '/scripts', as an alias for including the file straight from the node module.

dannycoates:
This looks good to merge. A next step might be to use [envify](https://www.npmjs.com/package/envify) to get rid of the template to set the dsn.

PR #116, opened 2017-06-22
add /__lbheartbeat__ endpoint

relud:
@dannycoates This is needed to comply with [Dockerflow](https://github.com/mozilla-services/Dockerflow).


dannycoates:
thanks @relud

PR #132, opened 2017-06-23
Add /__version__ route

### LINKED ISSUE #33: Add /__version__ route?
Re: https://p2p.dev.lcip.org

Speaking of interesting challenges, creating a `/__version__` route which lets us (friendly neighborhood QA) know exactly what code (Git SHA) is deployed on the dev/stage/prod server and would be pretty useful [eventually].

Currently https://p2p.dev.lcip.org/__version__ returns a 404 (as expected).

Here's a sample of what we get back from something like TestPilot prod server (https://testpilot.firefox.com/__version__):

```json
{
  "commit": "c4639a25b938bd4717aadc3ead840e80f0a0f059",
  "version": "2017-05-04",
  "source": "https://github.com/mozilla/testpilot"
}
```

pdehaan:
Fixes #33


pdehaan:
http://localhost:1443/__version__

```json
{
  "commit": "2691dfc",
  "source": "https://github.com/mozilla/something-awesome/",
  "version": "1.0.0"
}
```

But not sure if this will need tweaking for stage/prod deployments.

pdehaan:
Current feels:

pdehaan:
@dannycoates: Like this?

```js
// scripts/version.js
#!/usr/bin/env node

const fs = require('fs');
const path = require('path');
const pkg = require('../package.json');

let commit;

try {
  commit = require('git-rev-sync').short();
} catch (err) {
  // Whatever...
}

const filename = path.join(__dirname, '..', 'public', 'version.json');
const filedata = {
  commit,
  source: pkg.homepage,
  version: pkg.version
};

fs.writeFileSync(filename, JSON.stringify(filedata, null, 2) + "\n");
```

Then I guess the portal_server.js could handle the `/__version__` route by serving up the static version.json file:

```js
app.get('/__version__', (req, res) => {
  res.sendFile(path.join(STATIC_PATH, 'version.json'));
});
```

dannycoates:
> Like this?

Exactly! ❤️

package.json
```diff
@@ -12,6 +12,7 @@
     "cross-env": "^5.0.1",
     "express": "^4.15.3",
     "express-handlebars": "^3.0.0",
+    "git-rev-sync": "^1.9.1",
```
pdehaan: Wasn't sure if this should now be a dep or devdep.

package.json
```diff
@@ -38,12 +39,13 @@
   "engines": {
     "node": ">=8.0.0"
   },
+  "homepage": "https://github.com/mozilla/something-awesome/",
```
pdehaan: Added this because I'm using it for the `/__version__` endpoint's `source` field.

package.json
```diff
@@ -38,12 +39,13 @@
...
   "license": "MPL-2.0",
   "repository": "mozilla/something-awesome",
   "scripts": {
-    "bundle": "browserify frontend/src/main.js | uglifyjs > public/bundle.js",
+    "bundle": "browserify frontend/src/main.js | uglifyjs > public/bundle.js && node scripts/version",
```
pdehaan: Not sure about this part. Since `bundle` gets run as part of the `prestart` hook, it means any existing /public/version.json file would be overwritten when you start the server. So not sure if I'd be killing a file on stage/prod.
Or if we should decouple the `node scripts/version` bit and have people run it manually for local/dev server.

PR #168, opened 2017-07-06
Show error page if upload fails

dnarcese:
When a file cannot be uploaded, show a very simple error page (which will be changed with new designs).

### DIFF
--- a/frontend/src/fileSender.js
+++ b/frontend/src/fileSender.js
@@ -51,6 +51,9 @@ class FileSender extends EventEmitter {
         reader.onload = function(event) {
           resolve(new Uint8Array(this.result));
         };
+        reader.onerror = function(event) {
+          reject(event.explicitOriginalTarget.error.name);
+        };
       })
     ])
       .then(([secretKey, plaintext]) => {
--- a/frontend/src/fileSender.js
+++ b/frontend/src/fileSender.js
@@ -51,8 +51,8 @@ class FileSender extends EventEmitter {
         reader.onload = function(event) {
           resolve(new Uint8Array(this.result));
         };
-        reader.onerror = function(event) {
-          reject(event.explicitOriginalTarget.error.name);
+        reader.onerror = function(err) {
+          reject(err);
         };
       })
     ])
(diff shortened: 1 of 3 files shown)


frontend/src/fileSender.js
```diff
@@ -51,6 +51,9 @@ class FileSender extends EventEmitter {
         reader.onload = function(event) {
           resolve(new Uint8Array(this.result));
         };
+        reader.onerror = function(event) {
+          reject(event.explicitOriginalTarget.error.name);
```
pdehaan: @dnarcese @abhinadduri Are these errors logged and sent through Raven?

frontend/src/upload.js
```diff
@@ -66,35 +68,44 @@ $(document).ready(function() {
...
+        $('#file-list').hide();
+        $('#upload-progress').hide();
+        $('#share-link').show();
+        $('#upload-error').hide();
+
+        populateFileList(JSON.stringify(fileData));
+        notify('Your upload has finished.');
+      })
+      .catch(err => {
+        console.log('Upload error name: ' + err);
```
pdehaan: Same as above, are these errors logged and sent through Raven at all?

frontend/src/fileSender.js
abhinadduri: Nope, should be fairly easy to send them to Raven though.

frontend/src/upload.js
abhinadduri: This one is a bit more tricky: if the user accidentally tampers with the salt, this error would trigger. Is that something we would want to send to Raven?

frontend/src/fileSender.js
dnarcese: I think that only sending the error to Raven here is sufficient? This will eliminate any duplicates, since errors from the `reader.onerror` triggers (https://github.com/mozilla/something-awesome/pull/168#discussion-diff-126022292R54) are always caught here.

frontend/src/upload.js
dnarcese: I think that only sending the error to Raven here is sufficient? This will eliminate any duplicates, since errors from the `reader.onerror` triggers (https://github.com/mozilla/something-awesome/pull/168#discussion-diff-126022292R54) are always caught here.

PR #177, opened 2017-07-10
Gcm compliance

abhinadduri:


views/index.handlebars
```diff
@@ -100,6 +100,16 @@
       </div>
     </div>
 
+    <div id="compliance-error">
+      <div class="title">
+        Encryption error<br>
+        Your browser does not support gcm encryption.
+      </div>
+      <div class="send-new">
```
dannycoates: do we need this?

frontend/src/utils.js
```diff
@@ -32,8 +32,42 @@ function notify(str) {
...
+        length: 128
+      },
+      true,
+      ['encrypt', 'decrypt']
+    ).then(key => {
+      return window.crypto.subtle.encrypt(
+        {
+          name: 'AES-GCM',
+          iv: window.crypto.getRandomValues(new Uint8Array(12)),
+          additionalData: window.crypto.getRandomValues(new Uint8Array(6))
```
dannycoates: do we need `tagLength` for IE/Edge?

views/index.handlebars
abhinadduri: oops...

views/index.handlebars
```diff
@@ -100,6 +100,13 @@
       </div>
     </div>
 
+    <div id="compliance-error">
+      <div class="title">
+        Encryption error<br>
+        Your browser does not support gcm encryption.
```
pdehaan: Which browser/versions would this apply to?

It feels like this should be slightly more actionable (like prompting the user to the Download Firefox page or something), but maybe that is contained in the new UI specs.

abhinadduri: This applies to the current stable version of Safari.

Yeah, I left this fairly basic since it will eventually be replaced by the new UI.

PR #193, opened 2017-07-13
Frontend tests

### LINKED ISSUE #181: Unit test fileSender and fileReceiver

abhinadduri:
Fixes #181


package.json
```diff
@@ -52,6 +52,7 @@
     "lint:js": "eslint .",
     "start": "node server/server",
     "test": "mocha test/unit && mocha test/server",
+    "test-browser": "watchify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d",
```
pdehaan: if "test/frontend/bundle.js" is compiled/browserified, we may want to add it to .gitignore and nuke the file below.

abhinadduri: Ahh good idea. Did not mean to push that.

test/frontend/frontend.bundle.js
```diff
@@ -0,0 +1,22 @@
+class FakeFile extends Blob {
+    constructor(name, data, opt) {
+        super(data, opt);
+        this.name = name;
+    }
```
pdehaan: Indenting looks weird in this file (4 space indent). We may want to confirm that ESLint and prettier are linting these files. I would have almost expected CircleCI linting to fail on these.

pdehaan: This is the line in package.json that you may want to tweak:

```js
"format": "prettier '{frontend/src/,scripts/,server/,test/}*.js' 'public/*.css' --single-quote --write",
```

Instead of "test/", we'll need "test/**/" due to the nested files in the test directory.
Although, I _think_ if I currently run `npm run format` on the codebase it may break linting, so I'll shut up now.

test/frontend/frontend.test.js
```diff
@@ -0,0 +1,353 @@
...
+    return fr.download().then(([decrypted, name]) => {
+      assert(decrypted);
+      assert(name);
+      assert(!testHashing);
+    }).catch(err => {
+      assert.fail();
+    })
+  })
+
+  it('Should catch tampered checksums', function(done) {
```
dannycoates: This is testing the same thing as the next test, checksum mismatch, but in a more complicated way. I think we only need the simpler one. A good replacement would be a test that checks an secret key mismatch.

dannycoates: I think "tampered" isn't the best word for this. It took me a while to understand what you meant since tampering can happen at other times besides creation. Phony, bogus, or fraudulent better imply it's the creator that's being a jerk.

test/frontend/frontend.test.js
```diff
@@ -0,0 +1,353 @@
...
+        },
+        true,
+        ['encrypt', 'decrypt']
+      )
+      .then(key => {
+        return window.crypto.subtle.encrypt(
+          {
+            name: 'AES-GCM',
+            iv: hexToArray(encryptedIV),
+            additionalData: fileHash,
```
dannycoates: Add a comment here, this is the meat of this whole test, lets make it obvious that `fileHash` is the phony checksum

abhinadduri: Gotcha, will do.

frontend/src/fileReceiver.js
```diff
@@ -61,60 +61,51 @@ class FileReceiver extends EventEmitter {
...
+            resolve(decrypted);
           })
-        ]);
+        }),
+        new Promise((resolve, reject) => {
+          resolve(fdata.filename);
+        }),
+        new Promise((resolve, reject) => {
+          resolve(hexToArray(fdata.aad));
+        })
```
pdehaan: For these two promises, I don't think you need a Promise wrapper since you're doing a `Promise.all()` above. IIRC, you can just return `fdata.filename` and `hexToArray(fdata.aad)` and it should still work the same.

frontend/src/fileReceiver.js
```diff
@@ -61,60 +61,51 @@ class FileReceiver extends EventEmitter {
...
+          return Promise.reject();
+        } else {
+          this.emit('safe', true);
+          return Promise.all([
+            new Promise((resolve, reject) => {
+              resolve(decrypted);
+            }),
+            new Promise((resolve, reject) => {
+              resolve(fname);
+            })
```
pdehaan: Same here. I think you can just do `return Promise.all([decrypted, fname]);` or possibly even `Promise.resolve([decrypted, fname]);`.

circle.yml
```diff
@@ -5,6 +5,12 @@ machine:
     - docker
     - redis
 
+dependencies:
+  pre:
+    - npm i -g get-firefox
+    - get-firefox --platform linux --extract --target /home/ubuntu/send
+    - npm i -g geckodriver
```
pdehaan: I bet you can add this global install to the line 10 `npm i -g get-firefox geckodriver` and install them both at the same time.

Not sure if we'd want to add these as devDependencies if we'd ever need to run them locally. 🤷‍♀️

test/frontend/frontend.test.html
```diff
@@ -0,0 +1,27 @@
...
+  <script>
+    mocha.checkLeaks();
+    mocha.globals(['jQuery']);
+    mocha.run(function(err) {
+      document.title = err;
+    });
+  </script>
+
+  <script>
+  </script>
```
pdehaan: nit: _Probably_ don't need the empty `<script></script>` tag here.

frontend/src/fileReceiver.js
abhinadduri: Fixed

frontend/src/fileReceiver.js
abhinadduri: Fixed

test/frontend/frontend.test.html
abhinadduri: Fixed.

circle.yml
abhinadduri: @dannycoates What are your thoughts on adding these as devDependencies?

PR #197, opened 2017-07-13
fixes issues 195 and 192

### LINKED ISSUE #192: More bulletproof server meta validation
Randomly spotted while trying to track down why my CLI file uploading was failing...

Re: [/server/server.js:191-199](https://github.com/mozilla/send/blob/52173bf6e79c35fbba7e3493fae94ebf9d53b4c0/server/server.js#L191-L199), I think we're doing some things out of order here, since we're trying to validate the `meta.id` value before checking if the `meta` object has an 'id' property:

```js
  if (
    !validateIV(meta.id) ||
    !meta.hasOwnProperty('aad') ||
    !meta.hasOwnProperty('id') ||
    !meta.hasOwnProperty('filename')
  ) {
    res.sendStatus(404);
    return;
  }
```

### LINKED ISSUE #195: Investigate unicode filenames
Do we handle this already or not?

ref https://www.npmjs.com/package/content-disposition

abhinadduri:
Fixes #192 
Fixes #195


server/server.js
```diff
@@ -189,10 +189,10 @@ app.post('/upload', (req, res, next) => {
   }
 
   if (
-    !validateIV(meta.id) ||
     !meta.hasOwnProperty('aad') ||
     !meta.hasOwnProperty('id') ||
-    !meta.hasOwnProperty('filename')
+    !meta.hasOwnProperty('filename' ||
```
pdehaan: I think... this is wrong.

```
!meta.hasOwnProperty('filename') ||
!validateIV(meta.id)
```

abhinadduri: Oops, nice catch! Messed up some indenting and ended up with that.

frontend/src/download.js
```diff
@@ -65,7 +65,7 @@ $(document).ready(function() {
         return;
       })
       .then(([decrypted, fname]) => {
-        name.innerText = fname;
+        name.innerText = decodeURIComponent(fname);
```
abhinadduri: Yeah, will do.

PR #225, opened 2017-07-18
some character encoding fixes

abhinadduri:


frontend/src/utils.js
```diff
@@ -69,9 +69,25 @@ function gcmCompliant() {
   }
 }
 
+function escapeHtml(string) {
+  let map = {
```
pdehaan: Nit: use `const`

```
/home/ubuntu/send/frontend/src/utils.js
  73:7   error  'map' is never reassigned. Use 'const' instead  prefer-const
  84:42  error  Unnecessary escape character: \/                no-useless-escape
```

frontend/src/utils.js
```diff
@@ -69,9 +69,25 @@ function gcmCompliant() {
...
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
+    "'": '&#39;',
+    '/': '&#x2F;',
+    '`': '&#x60;',
+    '=': '&#x3D;'
+  };
+
+  return String(string).replace(/[&<>"'`=\/]/g, s => map[s]);
```
pdehaan:
```
/home/ubuntu/send/frontend/src/utils.js
  84:42  error  Unnecessary escape character: \/                no-useless-escape
```

dannycoates:
closing in favor of #232

PR #259, opened 2017-07-19
add alert when uploading multiple files

### LINKED ISSUE #247: There's no UI indication that the user can only upload one file

### LINKED ISSUE #248: Dragging two or more files onto the page appears to work but only uploads one file
There's no indication that this would occur.

## STR
* drag two files into send
* get download link
* download files

## Expected
* All files are downloaded

## Actual
* Only one file is downloaded

Fix: 
* the ultimate fix here is to handle multiple downloads, the immediate fix would be to not let people drag multiple files in and warn them about it.

dnarcese:
Fixes #247
Fixes #248

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -70,18 +70,32 @@ $(document).ready(function() {
   // on file upload by browse or drag & drop
   function onUpload(event) {
     event.preventDefault();
+    let file = '';
+    if (event.type === 'drop') {
+      if (event.originalEvent.dataTransfer.files.length > 1){
+        document.l10n.formatValue('uploadPageMultipleFilesAlert')
+                     .then(str => {
+                       alert(str);
+                     });
+        return;
+      }
+      file = event.originalEvent.dataTransfer.files[0];
+    } else {
+      if (event.target.files.length > 1){
+        document.l10n.formatValue('uploadPageMultipleFilesAlert')
+                     .then(str => {
+                       alert(str);
+                     });
+        return;
+      }
+      file = event.target.files[0];
+    }
+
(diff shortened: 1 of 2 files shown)


public/locales/send.en-US.ftl
```diff
@@ -4,10 +4,11 @@ uploadPageHeader = Private, Encrypted File Sharing
 uploadPageExplainer = Send files through a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
 
 uploadPageLearnMore = Learn more
-uploadPageDropMessage = Drop your files here to start uploading
+uploadPageDropMessage = Drop your file here to start uploading
```
dannycoates: Once we have other translations we'll need to change the key when we need to make changes to the string like this. For example if we were to change it back to "files" in v2 we'd rename the key to something like`uploadPageDropMultipleMessage` or `uploadPageDropMessage2`.

frontend/src/upload.js
```diff
@@ -70,18 +70,32 @@ $(document).ready(function() {
...
+      if (event.originalEvent.dataTransfer.files.length > 1){
+        document.l10n.formatValue('uploadPageMultipleFilesAlert')
+                     .then(str => {
+                       alert(str);
+                     });
+        return;
+      }
       file = event.originalEvent.dataTransfer.files[0];
     } else {
+      if (event.target.files.length > 1){
```
dannycoates: I don't think this case will ever happen as long as we don't have `multiple` on the input element, but there's no harm in checking here too.

dnarcese: That's true, I can remove that check.

PR #266, opened 2017-07-20
abort uploads over maxfilesize

### LINKED ISSUE #180: Set a max upload size on the server

dannycoates:
Fixes #180

* adds an alert error to the ui if the selected file's size is too big
* aborts uploads on the server that are too big
* adds the `bytes` module cuz im lazy

### DIFF
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -30,7 +30,7 @@ $(document).ready(function() {
       // update progress bar
       $('#dl-progress').circleProgress('value', percent);
       $('.percent-number').html(`${Math.floor(percent * 100)}`);
-      $('.progress-text').text(`${filename} (${bytes(progress[0])} of ${bytes(progress[1])})`);
+      $('.progress-text').text(`${filename} (${bytes(progress[0], {decimalPlaces: 1, fixedDecimals: true})} of ${bytes(progress[1], {decimalPlaces: 1})})`);
       //on complete
       if (percent === 1) {
         fileReceiver.removeAllListeners('progress');
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -1,5 +1,6 @@
 const FileReceiver = require('./fileReceiver');
 const { notify } = require('./utils');
+const bytes = require('bytes');
 const $ = require('jquery');
 require('jquery-circle-progress');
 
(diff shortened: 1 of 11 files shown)


johngruen:
R+ on the string

server/config.js
```diff
@@ -36,6 +36,11 @@ const conf = convict({
     format: ['production', 'development', 'test'],
     default: 'development',
     env: 'NODE_ENV'
+  },
+  max_file_size: {
+    format: Number,
+    default: (1024 * 1024 * 1024) * 2,
+    env: 'P2P_MAX_FILE_SIZE'
```
ericawright: I thought we were removing all "P2P" references - rename?

dannycoates: yeeeaaaah, i wanted to rename them all at once, but not in this PR, so... :man_shrugging:

ericawright: but it's a new variable, is it not? unless you have an .env file that I am unaware of?

server/config.js
```diff
@@ -36,6 +36,11 @@ const conf = convict({
     format: ['production', 'development', 'test'],
     default: 'development',
     env: 'NODE_ENV'
+  },
+  max_file_size: {
+    format: Number,
+    default: (1024 * 1024 * 1024) * 2,
```
pdehaan: Add a comment about what real-world value this is? I assume its 2GB.

dannycoates: comments always tend to diverge from reality. do the math 💪

frontend/src/download.js
```diff
@@ -29,20 +30,7 @@ $(document).ready(function() {
...
-      } else if (progress[1] < 1000000000) {
-        $('.progress-text').html(
-          `${filename} (${(progress[0] / 1000000).toFixed(1)}MB of ${(progress[1] / 1000000).toFixed(1)}MB)`
-        );
-      } else {
-        $('.progress-text').html(
-          `${filename} (${(progress[0] / 1000000).toFixed(1)}MB of ${(progress[1] / 1000000000).toFixed(1)}GB)`
-        );
-      }
+      $('.progress-text').text(`${filename} (${bytes(progress[0])} of ${bytes(progress[1])})`);
```
pdehaan: I think this may be causing some UI jank as well. You can see in the following UI screenshot that the filesizes are rounded to 2 decimal places. I think in some cases during upload, it's rounding the ".00" to ".0" which causes the UI to slightly shift very briefly.

This _may_ all be academic though since I believe @johngruen also has a PR to tweak the UI, so not sure who's PR will "win".

firefox_send

server/config.js
pdehaan: Fair enough. But to be clear, the UI says the recommended file limit is 1GB for uploads, but the server will allow up to ~2GB max?

firefox_send

dannycoates: both hard limits are the same. the recommendation is just a suggestion

Issue #260, opened 2017-07-19
Change /privacy and /terms links


wresuolc:
Right now these point to Test Pilot's legal pages, which is good, but the site also has legal pages, so they need to point to an intermediate page showing both.  That page should be hosted on send.f.c and the content should be:

> Firefox Send is currently a Test Pilot experiment, and subject to the Test Pilot [Terms of Service](https://testpilot.firefox.com/terms) and [Privacy Notice](https://testpilot.firefox.com/privacy). You can learn more about this experiment and its data collection here.
> 
> Use of the Firefox Send website is also subject to Mozilla’s [Websites Privacy Notice](https://www.mozilla.org/privacy/websites/) and [Websites Terms of Use](https://www.mozilla.org/about/legal/terms/mozilla/).

This is part of launch milestone.  /cc @ellee @johngruen

wresuolc:
Edit:  The "here" word should link to the page on Test Pilot.  @johngruen -- do we have a URL for it yet?

PR #268, opened 2017-07-20
Testpilot ga

abhinadduri:
Finished adding in all metrics.

One issue is that a user can click on `Try Firefox Send` without completing a download, something that we don't account for in our metrics guidelines for the `restarted` event.


frontend/src/upload.js
```diff
@@ -1,74 +1,142 @@
...
+    $('#link').attr('disabled', false);
+    $copyBtn.attr('data-l10n-id', 'copyUrlFormButton');
+
+    if (localStorage.length === 0) {
+      toggleHeader();
+    } else {
+      for (let i = 0; i < localStorage.length; i++) {
+        const id = localStorage.key(i);
+        //check if file exists before adding to list
+        checkExistence(id, true);
```
pdehaan: I'm a bit confused here... We're looping over `localStorage` and checking each key to see if the file exists. But now we're injecting more things into `localStorage`, such as `referrer`, etc.

Should we be filtering out `id`s against a regexp before calling `setExistence()`? Or should we possibly refactor the files into a specific key (such as `localStorage.files` which is an array/Object of all the non-expired files?

abhinadduri: Further down in the code, check existence does a check to ignore the injected elements. We can move that up here if necessary.

pdehaan: ... Because I notice we have code like this in ~4 places currently:

```diff
+          if (id !== 'totalUploads' && 
+              id !== 'totalDownloads' &&
+              id !== 'referrer') {
+            // ...
+          }
```

Maybe we need some utility `function isFile(id) {}` helper.

```js
function isFile(id) {
  return !['referrer', 'totalDownloads', 'totalUploads'].contains(id); // optionally validate `id` against a RegExp.
}

File: frontend/src/upload.js [hunk 2]
```diff
@@ -99,6 +167,18 @@ $(document).ready(function() {
...
+      localStorage.setItem('referrer', 'cancelled-upload');
+      
+      // record upload-stopped (cancelled) by sender
+      window.analytics
+            .sendEvent('sender', 'upload-stopped', {
+              cm1: file.size,
+              cm5: localStorage.getItem('totalUploads'),
+              cm6: unexpiredFiles,
+              cm7: totalDownloads,
+              cd1: event.type === 'drop' ? 'drop' : 'click',
```
pdehaan: Is there any way the `event.type` could not be "drop" or "click"?

abhinadduri: The only way for the `event.type` to not be `drop` is if the user doesn't drag and drop the file. I'm not sure the `onChange` property of the input tag can be changed with anything other than a click or a drag for a file upload.

frontend/src/upload.js
abhinadduri: This would be a handy function to have! I'll go ahead and add it into `utils.js`.

frontend/src/upload.js
```diff
@@ -346,18 +501,67 @@ $(document).ready(function() {
...
+    if (localStorage.hasOwnProperty('totalDownloads')) {
+      totalDownloads = localStorage.getItem('totalDownloads');
+    }
+
     // delete file
     $popupText.find('.del-file').click(e => {
       FileSender.delete(file.fileId, file.deleteToken).then(() => {
         $(e.target).parents('tr').remove();
-        localStorage.removeItem(file.fileId);
+        const timeToExpiry = 86400000 - (new Date().getTime() - file.creationDate.getTime());
```
pdehaan: Not sure if we want to convert the magic 86400000 number to a `const`. For example:
```js
const timeToExpiry = ONE_DAY_IN_MS - Date.now() - file.creationDate.getTime();
```

frontend/src/utils.js
```diff
@@ -69,9 +69,35 @@ function gcmCompliant() {
...
+    case 'https://www.mozilla.org/':
+      return 'mozilla';
+    case 'https://www.mozilla.org/about/legal':
+      return 'legal';
+    case 'https://testpilot.firefox.com/about':
+      return 'about';
+    case 'https://testpilot.firefox.com/privacy':
+      return 'privacy';
+    case 'https://testpilot.firefox.com/terms':
+      return 'terms';
```
pdehaan: Re: #260
Not sure what "privacy" and "terms" means here, or if we need to try clarifying to disambiguate between Send Privacy/Terms and TestPilot Privacy/Terms. Or not, I have no idea what I'm talking about.
/cc @chuckharmston @clouserw

chuckharmston: They're intended to be the footer links. If there are additional link destinations I missed, feel free to disambiguate or add to the list as appropriate as long as you update the docs; I pieced this together from the mocks.

frontend/src/upload.js
```diff
@@ -1,74 +1,142 @@
...
+    })
+
+    $('body').on('dragover', allowDrop).on('drop', onUpload);
+    // reset copy button
+    const $copyBtn = $('#copy-btn');
+    $copyBtn.attr('disabled', false);
+    $('#link').attr('disabled', false);
+    $copyBtn.attr('data-l10n-id', 'copyUrlFormButton');
+
+    if (localStorage.length === 0) {
```
dannycoates: with the new keys this check is no longer a valid way to determine whether to show the file list header

server/server.js
```diff
@@ -107,12 +107,18 @@ app.get('/download/:id', (req, res) => {
...
+        storage
+          .ttl(id)
+          .then(timeToExpiry => {
+            res.render('download', {
+              filename: decodeURIComponent(filename),
+              filesize: bytes(contentLength),
+              sizeInBytes: contentLength,
+              timeToExpiry: timeToExpiry,
+              trackerId: conf.analytics_id,
+              dsn: conf.sentry_id
```
dannycoates: trackerId and dsn should not need to be in this response anymore

frontend/src/upload.js
```diff
@@ -141,27 +221,80 @@ $(document).ready(function() {
...
     });
 
+    let uploadStart;
     fileSender.on('encrypting', isStillEncrypting => {
       // The file is being encrypted
       if (isStillEncrypting) {
         console.log('Encrypting');
       } else {
         console.log('Finished encrypting');
+        uploadStart = new Date().getTime();
```
dannycoates: `Date.now()` is the preferred way to do this

frontend/src/download.js
```diff
@@ -1,14 +1,53 @@
...
+    })
+
+    $('#expired-send-new').click(function() {
+      localStorage.setItem('referrer', 'errored-download');
+    })
+
+  }
+
   const filename = $('#dl-filename').html();
+  const bytelength = $('#dl-bytelength').html();
```
dannycoates: should be `.text()` and converted to Number. same for `timeToExpiry`

frontend/src/upload.js
```diff
@@ -141,27 +221,80 @@ $(document).ready(function() {
...
+    for (let i = 0; i < localStorage.length; i++) {
+      const id = localStorage.key(i);
+      if (isFile(id)) {
+          unexpiredFiles += 1;
+      }
+    }
+
+    let totalDownloads = 0;
+    if (localStorage.hasOwnProperty('totalDownloads')) {
+      totalDownloads = localStorage.getItem('totalDownloads');
```
dannycoates: we should be consistent with the type of `totalDownloads` and others. localStorage stores everything as a string so helper functions to convert for us will mean less repetition:

```js
function getTotalDownloads() {
  return Number(localStorage.getItem('totalDownloads'))
}
```

frontend/src/upload.js
```diff
@@ -200,7 +345,9 @@ $(document).ready(function() {
             populateFileList(localStorage.getItem(id));
           }
         } else if (xhr.status === 404) {
-          localStorage.removeItem(id);
+          if (isFile(id)) {
```
dannycoates: we shouldn't even be calling `checkExistence` on non-file ids. move this up a level

frontend/src/upload.js
abhinadduri: Fixed by adding in a new function in `utils.js`.

frontend/src/upload.js
abhinadduri: Changed all instances of `new Date().getTime()` to `Date.now()`.

frontend/src/upload.js
abhinadduri: `checkExistence` is no longer called if an id is not a file.

server/server.js
abhinadduri: Removed.

frontend/src/download.js
abhinadduri: Changed to `const bytelength = Number($('#dl-bytelength').text());`, and the same for `timeToExpiry`.

frontend/src/download.js
```diff
@@ -20,9 +61,31 @@ $(document).ready(function() {
...
   function download() {
+    storage.totalDownloads += 1;
+    
     const fileReceiver = new FileReceiver();
 
+    const unexpiredFiles = storage.numFiles;
+    let totalUploads = 0;
+    if (storage.has('totalUploads')) {
+      totalUploads = storage.totalUploads;
+    }
```
dannycoates: we've got two blocks with this pattern. this is initialization code so i suggest moving them into the `Storage` constructor. Also check that they're necessary at all because I think the getters (as implemented) will return 0 when it's not already set.

frontend/src/download.js
```diff
@@ -20,9 +61,31 @@ $(document).ready(function() {
...
+
     fileReceiver.on('progress', progress => {
+
+      window.onunload = function() {
+        storage.referrer = 'cancelled-download';
+        // record download-stopped (cancelled by tab close or reload)
+        window.analytics
+              .sendEvent('recipient', 'download-stopped', {
+                cm1: bytelength,
+                cm5: totalUploads,
```
dannycoates: nit: lets use `storage.totalUploads` directly and remove the local var

abhinadduri: Fixed in all cases.

frontend/src/download.js
abhinadduri: Fixed by removing the check.

frontend/src/storage.js
```diff
@@ -0,0 +1,66 @@
...
+  get numFiles() {
+    let length = 0;
+    for (let i = 0; i < this.engine.length; i++) {
+      const k = this.engine.key(i);
+      if (isFile(k)) {
+        length += 1;
+      }
+    }
+    return length;
+  }
```
pdehaan: This looks mostly similar to `get files() {...}` above (with the obvious exception of returning a count instead of an array). Not sure if it's worth it just to return `this.files.length`, or if that is too 'expensive' due to the `JSON.parse()` calls.

This would probably be a lot easier if we just had a localStorage key named `files` which was a JSON encoded array/object of files, instead of storing everything at the root level and having to loop/filter. 🤷‍♀️

views/download.handlebars
```diff
@@ -41,6 +43,6 @@
     <img src="/resources/illustration_expired.svg" id="expired-img" data-l10n-id="linkExpiredAlt"/>
   </div>
   <div class="expired-description" data-l10n-id="uploadPageExplainer"></div>
-  <a class="send-new" data-l10n-id="sendYourFilesLink"></a>
+  <a class="send-new" id = "expired-send-new" data-l10n-id="sendYourFilesLink"></a>
```
pdehaan: Nit: remove spaces around `=`.

frontend/src/storage.js
abhinadduri: I avoided calling length because of the parsing calls. I'm not sure if it's worth it or not to just go with `this.files.length`.

frontend/src/download.js
```diff
@@ -1,14 +1,53 @@
...
+const Storage = require('./storage');
+const storage = new Storage(localStorage);
 const $ = require('jquery');
 require('jquery-circle-progress');
 
 const Raven = window.Raven;
+
+if (!storage.has('totalDownloads')) {
+  storage.totalDownloads = 0;
+}
```
dannycoates: delete me

frontend/src/storage.js
dannycoates: we can revisit the internals of this later. for now im just glad it's factored into it's own module

Issue #57, opened 2017-06-07
Localization


dannycoates:
???

pdehaan:
Here's how Firefox Notes is doin' it: https://github.com/mozilla/notes/pull/96

dannycoates:
Added l20n in #245

wresuolc:
@flodolo This project is ready for adding to Pontoon we think

wresuolc:
https://github.com/mozilla/send/tree/master/public/locales is the directory.  is that an ok location or does it need to be in the root?

flodolo:
CC @mathjazz

I'm pretty sure locales doesn't need to be in the root, but that structure is wrong
https://developer.mozilla.org/en-US/docs/Mozilla/Implementing_Pontoon_in_a_Mozilla_website

en-US should be a folder, and disappear from the filename.

The whole file needs to be reviewed, and comments added. Not completely sure how to do that efficiently. Maybe remove the file completely, and open a PR with the correct structure and filename so I can comment?

Also, there are a ton more information I need before enabling a project in Pontoon:
* Deadlines
* Dev server and how to test
* Is "Firefox Send" a brand name? Can it be localized?
* Pontoon needs access to the repository

flodolo:
Also cc @stasm and @pike. A few notes from a quick look at the FTL file, to speed up the review later. As said, I think the easiest way is to remove the add the correct FTL in a pull request and use that to comment.

nit: use 4 spaces instead of 2 when indenting attributes.

Do you really need this?

```
uploadingPageCancel = Cancel upload
  .title = {uploadingPageCancel}
```

It should probably just be

```
uploadingPageCancel
    .title = Cancel upload
```

```
sentFilesTitle1 = File
sentFilesTitle2 = Copy URL
sentFilesTitle3 = Expires In
sentFilesTitle4 = Delete
```

These don't look like titles, let's find better string IDs.

This

```
deleteButtonHover.title = {sentFilesTitle4}
```

Should be

```
deleteButtonHover
    .title = Delete
```

Unless strictly necessarily, don't reuse strings like this.

Pike:
We'll also need a top-level ```l10n.toml``` file, similar to the one I've created for formautofill, https://github.com/mozilla-l10n/formautofill-l10n/pull/1.

Also, for now please make sure that your file passes through https://github.com/projectfluent/python-fluent/blob/master/tools/fluentfmt.py, 'cause otherwise our toolchains can create weird and buggy output.

[...]

PR #269, opened 2017-07-20
refactored ftl file

abhinadduri:


dannycoates:
ref #57

dannycoates:
@flodolo r?

public/locales/en-US/send.ftl
```diff
@@ -7,7 +7,7 @@ uploadPageLearnMore = Learn more
 uploadPageDropMessage = Drop your file here to start uploading
 uploadPageSizeMessage = For the most reliable operation, it’s best to keep your file under 1GB
 uploadPageBrowseButton = Select a file on your computer
```
flodolo: Do you actually need the string on the entity, or only on the title attribute? If it's the former, this should be

```
uploadPageBrowseButton
    .title = Select a file on your computer
```

Same question for similar strings.

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
 
 uploadingPageMessage = Once your file uploads you will be able to set expiry options.
 uploadingPageCancel = Cancel upload
-  .title = {uploadingPageCancel}
+  .title = Cancel upload
 uploadCancelNotification = Your upload was cancelled.
 
 uploadingPageLargeFileMessage = This file is large and may take awhile to upload. Sit tight!
```
flodolo: Pretty sure "awhile" -> "a while"

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
 
-uploadSvgAlt.alt = Upload
-
-
-//Note the spec suggests that this string is editable. That feature will not appear at Launch
+uploadSvgAlt
+    .alt = Upload
 uploadSuccessTimingHeader = The link to your file will expire after 1 download or in 24 hours.
 copyUrlFormLabel = Copy and share the link to send your file:
 copyUrlFormLabelWithName = Copy and share the link to send your file: { $filename }
```
flodolo: What's the difference between these two strings?

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
 
 deleteFileButton = Delete file
-  .title = {deleteFileButton}
+    .title = Delete file
 sendAnotherFileLink = Send another file
-  .title = {sendAnotherFileLink}
+    .title = Send another file
 
-downloadAltText.alt = Download
+downloadAltText
```
flodolo: Is this an alt text on a button or link? We need a comment to explain it's an action, not a noun

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
+    .alt = Download
 downloadFileName  = Download { $filename }
 downloadFileSize = ({ $size })
```
flodolo: Where does size come from and how it's formatted?

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
 downloadMessage = Your friend is sending you a file with Firefox Send, a service that allows you to share files with a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
 downloadButtonLabel = Download
-  .title = {downloadButtonLabel}
+    .title = Download
 downloadNotification = Your download has completed.
 downloadFinish = Download Complete
 
 sendYourFilesLink = Try Firefox Send
```
flodolo: Needs a note in case Firefox Send should not be localized

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
 errorPageMessage = There has been an error uploading the file.
 errorPageLink = Send another file
 
 
-linkExpiredAlt.alt = Link expired
+linkExpiredAlt
+    .alt = Link expired
 expiredPageHeader = This link has expired or never existed in the first place!
 notSupportedHeader = Your browser is not supported.
 notSupportedDetail = Unfortunately this browser does not support the web technology that powers Firefox Send. You'll need to try another browser. We recommend Firefox!
```
flodolo: Nit: proper apostrophe instead of straight quote.

Note about Firefox Send if needed.

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
 
 
-sentFilesTitle1 = File
-sentFilesTitle2 = Copy URL
-sentFilesTitle3 = Expires In
-sentFilesTitle4 = Delete
-
+uploadedFile = File
+copyFileList = Copy URL
+expiryFileList = Expires In
```
flodolo: How is this string used? It smells of contenation

public/locales/en-US/send.ftl
```diff
@@ -19,66 +19,70 @@ notifyUploadDone = Your upload has finished.
...
+deleteFileList = Delete
 nevermindButton = Nevermind
```
flodolo: "Never mind"?

public/locales/en-US/send.ftl
abhinadduri: We need the string on the entity as well.

flodolo: OK. @mathjazz is this going to work without issues in Pontoon?

abhinadduri:
@flodolo Is there any documentation on the l10n.toml file you're mentioning? I'm looking at the linked toml file and I'm not too sure how to translate it to Send. Thanks!

flodolo:
http://moz-l10n-config.readthedocs.io/en/latest/fileformat.html

It should be basically identical, with public/locales instead of locales, and I guess we can drop the [locales] section for now.

public/locales/en-US/send.ftl
abhinadduri: Both html snippets are the exact same, except one requires a filename. Is there a clean way to do this? Maybe we could pass in an empty string to the data-l10n-args?

flodolo: Can you show me a screenshot of these two strings? I'm not sure I understand how and when they would be used, e.g. the first has a colon because it's followed by something?

public/locales/en-US/send.ftl
abhinadduri: It's an alt text on an svg element. I'll go ahead and add in a comment.

public/locales/en-US/send.ftl
abhinadduri: The file size comes in from the server and is rendered through handlebars to the ftl file. It's a string in a human readable format, i.e. 12 Mb.

abhinadduri:
@flodolo I have added in the requested changes and the toml file. Please let me know if there are any more changes I need to make. Thanks.

public/locales/en-US/send.ftl
```diff
@@ -43,16 +44,21 @@ deleteFileButton = Delete file
...
 downloadFileSize = ({ $size })
 downloadMessage = Your friend is sending you a file with Firefox Send, a service that allows you to share files with a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
+
+// This is also an action.
 downloadButtonLabel = Download
     .title = Download
 downloadNotification = Your download has completed.
 downloadFinish = Download Complete
 
+// We might not need to localize this if it's a brand name.
```
flodolo: You need to answer this question and a comment with the answer ;-)

public/locales/en-US/send.ftl
```diff
@@ -43,16 +44,21 @@ deleteFileButton = Delete file
 sendAnotherFileLink = Send another file
     .title = Send another file
 
+// This alt text and the download filename string are actions.
```
flodolo: Alternative text used on the download link/button (indicates an action).

public/locales/en-US/send.ftl
```diff
@@ -43,16 +44,21 @@ deleteFileButton = Delete file
...
 downloadAltText
     .alt = Download
 downloadFileName  = Download { $filename }
+
 downloadFileSize = ({ $size })
 downloadMessage = Your friend is sending you a file with Firefox Send, a service that allows you to share files with a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
+
+// This is also an action.
```
flodolo: Alternative text used on the download link/button (indicates an action).

(localizers don't see the whole file, only one string at a time)

flodolo:
What about the question on copy/legal review? Also, as explained in a comment, you need to answer the question "Is Firefox Send a brand name and should not be localized".

public/locales/en-US/send.ftl
abhinadduri: Whoops... @wresuolc what are your thoughts on this?

public/locales/en-US/send.ftl
```diff
@@ -19,66 +20,76 @@ notifyUploadDone = Your upload has finished.
...
 downloadFirefoxButtonSub = Free Download
 
 
-sentFilesTitle1 = File
-sentFilesTitle2 = Copy URL
-sentFilesTitle3 = Expires In
-sentFilesTitle4 = Delete
+uploadedFile = File
+copyFileList = Copy URL
+expiryFileList = Expires In
```
flodolo: What is this string followed by?

public/locales/en-US/send.ftl
abhinadduri: `copyUrlFormLabel` is only being used in one place. I refactored that line so we can safely remove this string now.

public/locales/en-US/send.ftl
abhinadduri: The `Expires In` is a column header. It isn't followed by anything, but there are times listed below it (one for each file).

flodolo: OK, can you add a comment explaining that? As simple as

// expiryFileList is used as a column header

P.S. I'm in EU timezone, I'll probably catch up next week with this

abhinadduri: Sure thing. Thanks for all the suggestions!

public/locales/en-US/send.ftl
```diff
@@ -19,68 +20,82 @@ notifyUploadDone = Your upload has finished.
 
 uploadingPageMessage = Once your file uploads you will be able to set expiry options.
 uploadingPageCancel = Cancel upload
-  .title = {uploadingPageCancel}
+  .title = Cancel upload
```
flodolo: 4 spaces indentation

public/locales/en-US/send.ftl
```diff
@@ -19,68 +20,82 @@ notifyUploadDone = Your upload has finished.
...
+// Note: Title text for button should be the same.
 sendAnotherFileLink = Send another file
-  .title = {sendAnotherFileLink}
+    .title = Send another file
 
-downloadAltText.alt = Download
+// Alternative text used on the download link/button (indicates an action).
+downloadAltText
+    .alt = Download
 downloadFileName  = Download { $filename }
```
flodolo: One space before =

Pike:
> @Pike do we care about blank lines between strings?

I'm afraid we do right now.

flodolo:
A few more questions that I need answered before opening up the project on Pontoon:
* Is there a stage/dev server? Will it be updated automatically with updated localizations? How often?
* Are there deadlines?
* How often will production be updated with new localizations?

abhinadduri:
@dannycoates Any idea on localization deadlines and update times?

abhinadduri:
@flodolo So should I remove all line breaks from the ftl file?

public/locales/en-US/send.ftl
abhinadduri: Fixed.

public/locales/en-US/send.ftl
abhinadduri: Fixed, thanks!

flodolo:
> @flodolo So should I remove all line breaks from the ftl file?

Empty lines, yes. You can try running the script above on your FTL file
https://github.com/projectfluent/python-fluent/blob/master/tools/fluentfmt.py

dannycoates:
We have no deadlines since we can deploy as translations are completed.

dev site: https://send.dev.mozaws.net (updated automatically on each push to master)
stage site: https://send.stage.mozaws.net (updated automatically on version tag)
production site: https://send.firefox.com (updated manually after stage)

We're able to deploy translations to production at any time, but likely with new versions once a week.

flodolo:
Can you also add write access to https://github.com/mozilla-pontoon? 

Once the file has been updated and merged, I'll try setting up Pontoon on stage tomorrow and check that everything works as expected.

dannycoates:
> Can you also add write access to https://github.com/mozilla-pontoon?

done 👍

abhinadduri:
@flodolo I removed the empty lines and am running the script, which seems to be outputting the same ftl as the input file. However, there's also an extra newline at the end of the python program's output: will this be an issue?

flodolo:
Let's keep the same output, so empty line at the end of the file.

flodolo:
One more note: once we start localizing, rule for string changes is the same as Firefox
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

public/locales/en-US/send.ftl
```diff
@@ -1,98 +1,94 @@
...
-errorAltText.alt = Upload error
+errorAltText
+    .alt = Upload error
 errorPageHeader = Something went wrong!
 errorPageMessage = There has been an error uploading the file.
 errorPageLink = Send another file
-
 fileTooBig = That file is too big to upload. It should be less than { $size }.
-
 linkExpiredAlt.alt = Link expired
```
flodolo: This need to be updated

abhinadduri: Just fixed this!

abhinadduri:
@flodolo I've added in the requested changes above. The latest push has an ftl file that matches the output of running `fluentfmt.py` on the file. Please let me know if there are more changes that need to be made before this merges into maser. Thanks so much!

PR #281, opened 2017-07-23
Stop ESLint from linting the /public/ directory

pdehaan:
The /public/ directory is an odd mix of generated and static files. We don't need to ESLint anything in there, and it was blowing up my computer because of minified l20n files that were being copied from node_modules, and an old bundle file.

### DIFF
--- a/test/.eslintrc.yml
+++ b/test/.eslintrc.yml
@@ -19,3 +19,5 @@ rules:
   mocha/no-pending-tests: error
   mocha/no-return-and-callback: warn
   mocha/no-skipped-tests: error
+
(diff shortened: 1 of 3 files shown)


test/.eslintrc.yml
```diff
@@ -19,3 +19,5 @@ rules:
   mocha/no-pending-tests: error
   mocha/no-return-and-callback: warn
   mocha/no-skipped-tests: error
+
+  no-console: off # ¯\_(ツ)_/¯
```
pdehaan: Ref #283

dannycoates:
> because of minified l20n files that were being copied from node_modules

whoops, my fault, sorry

PR #295, opened 2017-07-24
renamed environment variables to remove P2P_ prefix

dannycoates:
@relud I renamed the environment variables. cool? These will need to change in the deployment


pdehaan:
@dannycoates What about [/docker-compose.yml:10](https://github.com/mozilla/send/blob/fa1f0208a4a5b2e6dbc04f47b2a132e679daf3e0/docker-compose.yml#L10)?

Issue #309, opened 2017-07-25
Expired link still accessible


ericawright:
Possibly related to  #159

STR: 
- Copy a link that is about to expire
- Visit the page after the time it should have expired

EXPECTED: 
- the link should display the expired page

ACTUAL:
- the link is still valid
- the file can still be downloaded!

Does the file actually get deleted from the server at 24 hours?

wresuolc:
please double check!

ericawright:
currently this is what we see from an expired link
![screen shot 2017-07-25 at 2 31 59 pm](https://user-images.githubusercontent.com/10803178/28587740-310ff6fa-7146-11e7-88c0-c3548f46af69.png)
and when clicking on that expired link (after #310), we see a flash, then the same thing

Issue #159, opened 2017-07-04
[Mac OS] Download option still displayed for expired file after browser restart


SoftVision-CiprianMuresan:
[Notes]:
- A similar behavior can be observed if the browser is restarted right after downloading a file. Instead of the 404 page, the download card will be displayed.

[Affected versions]:
- Firefox 54.0 and up

[Affected Platforms]:
- All Mac

[Prerequisites]:
- "When Firefox starts: Show my windows and tabs from last time" option is set in about:preferences

[Steps to reproduce]:
1. Start the Firefox browser with the profile from prerequisites and navigate to "https://send.dev.mozaws.net/" page.
2. Upload a file and open the download link in two different tabs.
3. Download the file in the first tab and then navigate to the second tab.
4. Close the browser (Quit) and reopen it with the same profile.
5. Observe the displayed tab.

[Expected result]:
- The "Link expired" page is displayed.

[Actual results]:
- The Download card is displayed.

[Additional notes]:
- This issue is also reproducible on the stage page (https://send.stage.mozaws.net). 
- If the page is manually refreshed, the "Link expired" page is correctly displayed.
- Here is a screen recording with the issue: [Link](https://drive.google.com/open?id=0B8ICzPJynbchSlI1MFVGQUlCUDA).

SoftVision-CiprianMuresan:
This issue is now reproducible on Windows and Linux too.

dannycoates:
fixed in #297

SoftVision-CiprianMuresan:
This issue is still reproducible on all platforms. Moreover after clicking the "Download" button for an expired link the download page is somehow merged with the Link expired page.

<img width="1280" alt="screen shot 2017-07-25 at 16 08 24" src="https://user-images.githubusercontent.com/18304511/28573642-ec807e5c-7153-11e7-82fc-21c36824a02a.png">

Screen recording: [here](https://drive.google.com/open?id=0B8ICzPJynbchcGRVdG9OZmI0TWM).

dannycoates:
Any link created before about 2pm PST yesterday (24th) will have this behavior

dannycoates:
The weird mixed error/download screen is fixed in #310. 

Having the page show that the link is expired immediately after a browser restart is difficult because the page is served from the browser cache, so we'd need to make an xhr request to check the status, which seems excessive given its rarity.

PR #311, opened 2017-07-25
expired ids should reject instead of returning null

dannycoates:
addresses the null problem in #309


abhinadduri:
Looks good!

ericawright:
I think all of the odd behaviour in #309 and #159 have been fixed now!

PR #314, opened 2017-07-25
added L10N_DEV environment variable for making all languages available

### LINKED ISSUE #312: expand availableLanguages in dev
To preview language changes on https://send.dev.mozaws.net we need to make the `availableLanguages` meta tag include all the locales in /public/locales

testpilot does this with a build script https://github.com/mozilla/testpilot/blob/a9ecec58c2e316ba85634af0aa18bace4757816d/frontend/tasks/pages.js#L119

dannycoates:
fixes #312 

@relud can we set `L10N_DEV=true` for send.dev.mozaws.net only?

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -9,6 +9,7 @@ const conf = require('./config.js');
 const storage = require('./storage.js');
 const Raven = require('raven');
 const crypto = require('crypto');
+const fs = require('fs');
 
 if (conf.sentry_dsn) {
   Raven.config(conf.sentry_dsn).install();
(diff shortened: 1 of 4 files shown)


relud:
sure. does `L10N_DEV=false` suffice for non-dev deployments?

relud:
config added, will be present for next deploy to dev

server/server.js
```diff
@@ -22,11 +23,34 @@ const STATIC_PATH = path.join(__dirname, '../public');
...
+
 app.engine(
   'handlebars',
   exphbs({
     defaultLayout: 'main',
-    partialsDir: 'views/partials/'
+    partialsDir: 'views/partials/',
+    helpers: {
+      availableLanguages,
+      l10nDev: conf.l10n_dev
```
pdehaan: Do we use this helper, or just the `availableLanguages` helper?

dannycoates: No, I was going to use it to add pontoon.js but the CSP and other changes scared me away for now

pdehaan:
It was a bit tricky to get this running locally.
Not sure if we want to force the env var when running `npm run dev` locally:

```js
"dev": "npm run build && L10N_DEV=true npm start",
```

Although if we start passing env vars, I think we have to  use [**cross-env**](http://npm.im/cross-env) so it plays nice w/ Windows.

pdehaan:
But I'm all :100: for landing this, as I am seeing a couple issues w/ German when hacking locally:

firefox_send

PR #318, opened 2017-07-26
added pontoon.js when L10N_DEV=true

dannycoates:
Let's see if this works!

cc @flodolo

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -49,39 +49,58 @@ app.engine(
     partialsDir: 'views/partials/',
     helpers: {
       availableLanguages,
-      l10nDev: conf.l10n_dev
+      maybePontoon: function() {
+        return conf.l10n_dev
+          ? '<script src="https://pontoon.mozilla.org/pontoon.js"></script>'
+          : null;
+      }
     }
   })
 );
 app.set('view engine', 'handlebars');
 
-app.use(helmet());
+app.use(
+  helmet({
+    frameguard: {
+      action: 'allow-from',
+      domain: 'https://pontoon.mozilla.org'
+    }
+  })
+);
 app.use(
   helmet.hsts({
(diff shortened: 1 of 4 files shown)


pdehaan:
Unrelated to specific issue, but something funny happening w/ package-lock.json diffs in these recent PRs.
Not sure if we're all on different versions of npm@5, but we're seeing a +/-2000 line change on every odd commit.

server/server.js
```diff
@@ -49,39 +49,58 @@ app.engine(
...
+    frameAncestors: ["'none'"],
+    objectSrc: ["'none'"],
+    reportUri: '/__cspreport__'
+  }
+};
+
+if (conf.l10n_dev) {
+  const d = cspPolicy.directives;
+  d.scriptSrc.push('https://pontoon.mozilla.org');
+  d.imgSrc.push('https://pontoon.mozilla.org');
```
pdehaan: Do we load images/styles from pontoon?

flodolo:
You want to talk with @mathjazz for this, I'm not familiar with the process to set up in-content localization.

mathjazz:
Sadly, I'm not exactly sure this will add a lot of value to the l10n process. Pontoon doesn't offer anything ATM to navigate between different page views, except for links to subpages. Since this is a SPA, the users would only have an easy access to the homepage.

If you still want to give it a shot, I suggest we use the stage server. In which case all the URLs in the [docs (Section B)](https://developer.mozilla.org/en-US/docs/Mozilla/Implementing_Pontoon_in_a_Mozilla_website) need to be replaced with the stage URL - mozilla-pontoon-staging.herokuapp.com (which includes the path to the `pontoon.js` file in this patch).

dannycoates:
ok, let's table this for now

PR #323, opened 2017-07-26
disable upload/download notifications

dannycoates:


frontend/src/upload.js
```diff
@@ -180,10 +180,6 @@ $(document).ready(function() {
     const fileSender = new FileSender(file);
     $('#cancel-upload').click(() => {
       fileSender.cancel();
-      location.reload();
-      document.l10n.formatValue('uploadCancelNotification').then(str => {
-        notify(str);
```
dannycoates: this wasn't actually being shown anyway because of the reload

PR #327, opened 2017-07-27
Modify popup delete dialog

### LINKED ISSUE #326: [Send] Measurement for Delete dialog
Here is the link~
https://mozilla.github.io/testpilot-assets/Send/Send_13_delete_dialogue/
Thx

youwenliang:
Fixes #326 
Also, expand feedback button by default in mobile view since there's no hover state.

### DIFF
--- a/public/main.css
+++ b/public/main.css
@@ -275,33 +275,39 @@ tbody {
 /* The actual popup (appears on top) */
 .popup .popuptext {
   visibility: hidden;
-  min-width: 115px;
+  min-width: 204px;
+  min-height: 105px;
   background-color: #fff;
   color: #000;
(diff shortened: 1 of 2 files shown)


public/main.css
```diff
@@ -694,6 +733,18 @@ tbody {
...
+  .popup .popuptext {
+    left: auto;
+    right: -40px;
+  }
+
+  /* Popup arrow */
+  .popup .popuptext::after {
+    left: auto;
+    right: 36px;
+  }
```
dnarcese: I think the popup should move to the left earlier. It causes horizontal scrolling at any screen between widths 768px and 920px.

youwenliang: sounds good to me!

PR #337, opened 2017-07-27
support v 50 and 51 by not allowing const in loops

ericawright:
the 
`  prefer-const: error` 
rule in eslint forces `const` in for loops, But firefox 50 and 51 break when this is enforced. Since Test pilot supports v50+ I think we should remove this from our code and lint rules. 
(I could not find a eslint rule related to this that would allow `let` in loops, so have removed it altogether)
re: #334

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -80,6 +80,7 @@ $(document).ready(function() {
   if (files.length === 0) {
     toggleHeader();
   } else {
+    // eslint-disable-next-line prefer-const
     for (let index in files) {
       const id = files[index].fileId;
       //check if file still exists before adding to list
--- a/frontend/src/utils.js
+++ b/frontend/src/utils.js
@@ -1,5 +1,6 @@
 function arrayToHex(iv) {
   let hexStr = '';
+  // eslint-disable-next-line prefer-const
   for (let i in iv) {
     if (iv[i] < 16) {
       hexStr += '0' + iv[i].toString(16);


pdehaan:
> (I could not find a eslint rule related to this that would allow let in loops, so have removed it altogether)

Yeah, I don't think I ever found an ESLint flag/rule/plugin to allow `let` in a loop (which I think was a Firefox bug, so makes sense that it may not be supported). In other projects I think we usually resort to `// eslint-disable-next line no-const` in the few places we use it.

ericawright:
whoops, forgot about that eslint ability, thanks will do

frontend/src/upload.js
```diff
@@ -80,6 +80,7 @@ $(document).ready(function() {
   if (files.length === 0) {
     toggleHeader();
   } else {
+    // eslint-disable-next line no-const
```
pdehaan: Sorry, I typoed in my original message. Should be `eslint-disable-next-line`. 😞 

http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

ericawright:
yeah, working on that. it's being funny

frontend/src/utils.js
```diff
@@ -1,6 +1,6 @@
 function arrayToHex(iv) {
   let hexStr = '';
-  for (let i in iv) {
+  for (let i in iv) { // eslint-disable-line prefer-const
```
pdehaan: But possibly same  here... Try running `npm run format` to let **prettier** reformat everything, and I'd guess that it'll move the comment to the following line and you'd get ESLint errors if you ran `npm run lint`.

👍 to using the same `// eslint-disable-next-line prefer-const` as above, and moving it up to line 2.5 above.

frontend/src/utils.js
```diff
@@ -1,6 +1,7 @@
 function arrayToHex(iv) {
   let hexStr = '';
   for (let i in iv) {
+    // eslint-disable-next-line prefer-const
```
pdehaan: 👍, but move it above the `for..in` loop.

ericawright: omg what am I doing!?

ericawright:
OK I think this time it is safe. gah

pdehaan:
You're really making me earn my paycheck on this one...

Issue #335, opened 2017-07-27
en-US send.ftl file has duplicate messages


Pike:
```shell
compare-locales l10n.toml . en-US
en-US/send.ftl
    WARNING: copyUrlHover occurs 2 times
    WARNING: deleteButtonHover occurs 2 times
    ERROR: copyUrlHover occurs 2 times
    ERROR: deleteButtonHover occurs 2 times
```

The duplicated entries should be removed.

CC @flodolo @mathjazz @TheoChevalier

flodolo:
Fixing.

pdehaan:
Wow, that `compare-locales` is pretty awesome.
What'd be the best place to file an enhancement request? It seems like even if I get a "WARNING" or "ERROR" it still exits with a status code of 0.

```sh
$ compare-locales l10n.toml . en-US fr de

de/send.ftl
    WARNING: copyUrlHover occurs 2 times
    WARNING: deleteButtonHover occurs 2 times
    ERROR: copyUrlHover occurs 2 times
    ERROR: deleteButtonHover occurs 2 times
    ERROR: Missing value at line 10, column 1 for uploadPageBrowseButton
    ERROR: Missing attribute: title at line 35, column 1 for deleteFileButton

en-US/send.ftl
    WARNING: copyUrlHover occurs 2 times
    WARNING: deleteButtonHover occurs 2 times
    ERROR: copyUrlHover occurs 2 times
    ERROR: deleteButtonHover occurs 2 times

fr/send.ftl
    // add and localize this file

de:
changed: 64
changed_w: 384
errors: 4
unchanged: 4
unchanged_w: 6
warnings: 2
94% of entries changed

en-US:
errors: 2
unchanged: 68
unchanged_w: 390
warnings: 2
0% of entries changed

fr:
missingInFiles: 70
missing_w: 393
0% of entries changed


$ echo $? # 0
```

I was thinking it may be nice to try running compare-locales in CI as part of our linting (although not sure how challenging it'd be to install via `easy_install` on Circle-CI).

flodolo:
@pdehaan I'm pretty sure you're looking at an old version of the repo.

There are warning about duplicated strings in locales, but we should be able to ignore them (first update to the file will clean them up). I fixed all other errors.

pdehaan:
@flodolo Yeah, sorry, I didn't sync to master since I wanted to check the error codes. I was just commenting on how having a non-zero status code from `compare-locales` would be useful since it'd fail builds on any errors.

[...]

PR #338, opened 2017-07-27
Remove duplicated strings in en-US, fix nn-NO file

flodolo:


flodolo:
Fixes #335

public/locales/en-US/send.ftl
```diff
@@ -74,10 +74,6 @@ copyFileList = Copy URL
 expiryFileList = Expires In
 deleteFileList = Delete
 nevermindButton = Never mind
-deleteButtonHover
-    .title = Delete
-copyUrlHover
-    .title = Copy URL
```
flodolo: For reference, same strings appear on line 87

Issue #319, opened 2017-07-26
"Browser not supported" page is not displayed for older Firefox versions


SoftVision-CiprianMuresan:
[Notes]:
- We should also show this page for Internet Explorer as long as we plan on releasing without support for it.
- Seeing as Test Pilot is not supported on Firefox 49 and lower, we should show this page on these versions.
- Opening the page on Safari correctly shows the "Browser not supported" page.

[Affected versions]:
- Firefox 49.0 and down

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Steps to reproduce]:
1. Open the an old version of Firefox (49 or lower) and navigate to https://send.dev.mozaws.net/.
2. Observe the page.

[Expected result]:
- "Browser not supported" page is displayed.

[Actual results]:
- The page's UI is broken.

[Additional notes]:
- Here is a screen recording with the issue:
![browser not supported](https://user-images.githubusercontent.com/18304511/28623859-dfaa567c-721f-11e7-98a8-28c7443dcd92.gif)

SoftVision-CosminMuntean:
It seems that we have a regression here. This issue is now reproducible on dev and stage server. 
The "Browser not supported" page is no longer displayed on Firefox 49 and below. However, the UI is correctly displayed on older versions of Firefox (<49), but the upload/download is not working. 

Here is screen recording with the issue:
![browser not supported](https://user-images.githubusercontent.com/19226534/30376522-6039680a-9895-11e7-968b-7bf4ed7ce6cd.gif)

SoftVision-CosminMuntean:
This issue is still reproducible on all Send servers, but the behavior is changed. If you navigate to the Send page, the "https://send.stage.mozaws.net/unsupported/outdated" link is correctly displayed but the "Browser not supported" page is displayed only if the page is refreshed.

Here is a screen recording with the issue:
![browser unsuported at refresh](https://user-images.githubusercontent.com/19226534/31503118-75878964-af77-11e7-9a66-b37facee31f5.gif)

johngruen:
Boy, this feels wontfixy

johngruen:
will look at metrics

Issue #127, opened 2017-06-23
The File Share page is wrongly rendered and upload/download functionality is not working on IE and Edge


SoftVision-CosminMuntean:
[Affected versions]:
- Internet Explorer 11.0.9600.18697
- Edge 25.10586.672.0

[Affected Platforms]:
- All Windows

[Steps to reproduce]:
1. Start the IE/Edge browser navigate to "https://p2p.dev.lcip.org/" page.
2. Observe the page and verify if the upload/download functionality is working.

[Expected result]:
- The page is correctly rendered and the upload/download functionality is working as expected.

[Actual results]:
- The page is wrongly rendered and the upload/download is not working.

[Notes]:
- This issue is also reproducible on the stage page (https://fileshare.stage.mozaws.net). 
- Here is a screen shot with the issue:
Sender:
![sneder](https://user-images.githubusercontent.com/19226534/27484783-2d061c50-5833-11e7-83ad-115e4a3b657b.png)

Receiver:
![reciver](https://user-images.githubusercontent.com/19226534/27484776-27bf9e10-5833-11e7-8c8f-824453192306.png)

wresuolc:
This will probably be fixed with new designs

dannycoates:
The page seems to render fine now. The extremely slow crypto performance is another issue...

SoftVision-CosminMuntean:
I have retested this issue on multiple versions of Windows 10 and it looks that on older versions of Edge, the website is not correctly rendered. However, all tested Edge versions do not display the "Send" background image.

Here are the results:
- Windows 10 Creators Edition with Microsoft Edge v.40.15063.0.0
The website is **correctly rendered**
- Windows 10 Anniversary Edition with Microsoft Edge v.38.14393.1066.0 
The website is **correctly rendered**
- Windows 10 November Edition with Microsoft Edge v.25.10586.672.0 
The website is **wrongly rendered**. Here are two screenshots with the issue:
1. https://goo.gl/cb1RZv
2. https://goo.gl/qiUW2q

Also, on **Internet Explorer** the website is **wrongly rendered** on all versions of Windows.

ericawright:
perhaps for launch at least we exclude internet explorer?

dannycoates:
We're considering IE as unsupported for now. Edge is working in v1.1.0

PR #339, opened 2017-07-27
Show error page on firefox v49 and below

ericawright:
@abhinadduri can you create an appropriate GA event for this?

Everyone else, any opinions on what should be shown on this page?

/unsupported/outdated now looks like this:
screen shot 2017-07-27 at 2 21 43 pm

very similar to /unsupported/gcm 

link goes to "https://support.mozilla.org/kb/update-firefox-latest-version" which offers advice on how to update firefox

Fixes: #319.
Ref: #127.

### DIFF
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -9,3 +9,24 @@ window.analytics = new testPilotGA({
   ds: 'web',
   tid: window.trackerId
 });
+
+const isSender = location.pathname.includes('/download');
+
+gcmCompliant().catch(err => {
+  $('#page-one').attr('hidden', true);
+  $('#download').attr('hidden', true);
+  sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
+    cd6: err
+  }).then(() => {
+    location.replace('/unsupported');
+  });
+});
+
+if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
+    parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
+    sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
+      cd6: new Error('Firefox is outdated.')
+    }).then(() => {
+      location.replace('/unsupported');
+    });
+}
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -10,11 +10,9 @@ window.analytics = new testPilotGA({
   tid: window.trackerId
 });
 
-const isSender = location.pathname.includes('/download');
+const isSender = !location.pathname.includes('/download');
 
 gcmCompliant().catch(err => {
-  $('#page-one').attr('hidden', true);
-  $('#download').attr('hidden', true);
   sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
     cd6: err
   }).then(() => {
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -3,8 +3,24 @@ window.Raven.config(window.dsn).install();
 window.dsn = undefined;
 
 const testPilotGA = require('testpilot-ga');
+const {gcmCompliant, sendEvent} = require('./utils');
 window.analytics = new testPilotGA({
   an: 'Firefox Send',
   ds: 'web',
   tid: window.trackerId
 });
+
+gcmCompliant().catch(err => {
+  $('#page-one').attr('hidden', true);
+  $('#download').attr('hidden', true);
+  sendEvent('sender', 'unsupported', {
+    cd6: err
+  }).then(() => {
+    location.replace('/unsupported');
+  });
+});
+
+if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
+    parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
+    location.replace('/unsupported');
+}
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -16,7 +16,7 @@ gcmCompliant().catch(err => {
   sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
     cd6: err
   }).then(() => {
-    location.replace('/unsupported');
+    location.replace('/unsupported/gcm');
   });
 });
 
@@ -25,6 +25,6 @@ if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
     sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
       cd6: new Error('Firefox is outdated.')
     }).then(() => {
-      location.replace('/unsupported');
+      location.replace('/unsupported/outdated');
     });
 }
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -22,5 +22,9 @@ gcmCompliant().catch(err => {
 
 if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
     parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
-    location.replace('/unsupported');
+    sendEvent('sender', 'unsupported', {
+      cd6: 'Unsupported Firefox'
+    }).then(() => {
+      location.replace('/unsupported');
+    });
 }
--- a/frontend/src/common.js
+++ b/frontend/src/common.js
@@ -9,22 +9,3 @@ window.analytics = new testPilotGA({
   ds: 'web',
   tid: window.trackerId
 });
-
-gcmCompliant().catch(err => {
-  $('#page-one').attr('hidden', true);
-  $('#download').attr('hidden', true);
-  sendEvent('sender', 'unsupported', {
-    cd6: err
-  }).then(() => {
-    location.replace('/unsupported');
-  });
-});
-
-if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
-    parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
-    sendEvent('sender', 'unsupported', {
-      cd6: 'Unsupported Firefox'
-    }).then(() => {
-      location.replace('/unsupported');
-    });
-}
--- a/public/locales/en-US/send.ftl
+++ b/public/locales/en-US/send.ftl
@@ -67,6 +67,8 @@ expiredPageHeader = This link has expired or never existed in the first place!
 notSupportedHeader = Your browser is not supported.
 // Firefox Send is a brand name and should not be localized.
 notSupportedDetail = Unfortunately this browser does not support the web technology that powers Firefox Send. You’ll need to try another browser. We recommend Firefox!
+notSupportedOutdatedDetail = Unfortunately this version of Firefox does not support the web technology that powers Firefox Send. You’ll need to update your browser.
+updateFirefox = Update Firefox
 downloadFirefoxButtonSub = Free Download
 uploadedFile = File
 copyFileList = Copy URL
--- a/server/server.js
+++ b/server/server.js
@@ -96,8 +96,11 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
-app.get('/unsupported', (req, res) => {
-  res.render('unsupported');
+app.get('/unsupported/:reason', (req, res) => {
+  const outdated = req.params.reason === 'outdated'? true : false;
+  res.render('unsupported', {
+    outdated: outdated
+  });
 });
 
 app.get('/legal', (req, res) => {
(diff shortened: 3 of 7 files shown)


frontend/src/common.js
```diff
@@ -3,8 +3,28 @@ window.Raven.config(window.dsn).install();
...
 window.analytics = new testPilotGA({
   an: 'Firefox Send',
   ds: 'web',
   tid: window.trackerId
 });
+
+gcmCompliant().catch(err => {
+  $('#page-one').attr('hidden', true);
+  $('#download').attr('hidden', true);
+  sendEvent('sender', 'unsupported', {
```
pdehaan: Below this was "recipient" in /frontend/src/download.js (but "sender" in the upload.js).
Is this fine here as "sender" always?

frontend/src/common.js
```diff
@@ -3,8 +3,28 @@ window.Raven.config(window.dsn).install();
...
+    cd6: err
+  }).then(() => {
+    location.replace('/unsupported');
+  });
+});
+
+if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1 &&
+    parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
+    sendEvent('sender', 'unsupported', {
+      cd6: 'Unsupported Firefox'
```
pdehaan: Is this fine as a string?
In every other instance of [`cd6` I grepped](https://github.com/mozilla/send/search?utf8=%E2%9C%93&q=cd6&type=), the `cd6` was an error object.

**Disclaimer:** I know nothing about metrics collection.

frontend/src/common.js
abhinadduri: No, it's not. Good catch, we might have to take this out of common.js.

ericawright: oh, yeah, didn't see that

abhinadduri: Like @pdehaan mentioned above, this metric can't always have the `sender` token. I'll move this back into upload and download and add the respective `sender` and `recipient` tags.

abhinadduri:
Do you think we should keep it in `common` and maybe look through the url for a `download` to determine if it's the `recipient` or the `sender`? Or do you think it would be better to move it out of `common` and into `upload` and `download`?

ericawright:
I like not having to repeat the code twice, but then it does add an extra step to figure out which page

abhinadduri:
I've added a check in common to address this, let me know what you think. Also @pdehaan, it seems like linting is breaking because we're not using `const` in for loops (but, doing this supposedly breaks older versions of Firefox).

pdehaan:
Semi related, but do we want to do similar UA sniffing and redirect IE users to /unsupported?
Ref: https://github.com/mozilla/send/issues/127#issuecomment-317456355

ericawright:
@pdehaan yeah, that's my plan at least

ericawright:
@SoftVision-CiprianMuresan can you test on IE for me to ensure that it is being redirected to the unsupported page?

public/locales/en-US/send.ftl
```diff
@@ -67,6 +67,8 @@ expiredPageHeader = This link has expired or never existed in the first place!
 notSupportedHeader = Your browser is not supported.
 // Firefox Send is a brand name and should not be localized.
 notSupportedDetail = Unfortunately this browser does not support the web technology that powers Firefox Send. You’ll need to try another browser. We recommend Firefox!
+notSupportedOutdatedDetail = Unfortunately this version of Firefox does not support the web technology that powers Firefox Send. You’ll need to update your browser.
+updateFirefox = Update Firefox
```
pdehaan: /cc @flodolo since this adds some l10n strings.

server/server.js
```diff
@@ -96,8 +96,11 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
-app.get('/unsupported', (req, res) => {
-  res.render('unsupported');
+app.get('/unsupported/:reason', (req, res) => {
+  const outdated = req.params.reason === 'outdated'? true : false;
```
pdehaan: I think the `? true : false` sorcery is unnecessary here since `req.params.reason === 'outdated'` should be Boolean-enough(tm).

ericawright: true

views/unsupported.handlebars
```diff
@@ -1,11 +1,19 @@
...
   <div class="title" data-l10n-id="notSupportedHeader"></div>
-  <div class="description" data-l10n-id="notSupportedDetail"></div>
-  <a id="dl-firefox" href="https://www.mozilla.org/firefox/new/?scene=2" target="_blank">
-    <img src="/resources/firefox_logo-only.svg" id="firefox-logo" alt="Firefox"/>
-    <div id="dl-firefox-text">Firefox<br>
-      <span data-l10n-id="downloadFirefoxButtonSub"></span>
-    </div>
-  </a>
+  {{#if outdated}}
+    <div class="description" data-l10n-id="notSupportedOutdatedDetail"></div>
```
pdehaan: Re: `notSupportedOutdatedDetail = Unfortunately this version of Firefox does not support the web technology that powers Firefox Send. You’ll need to update your browser.`

Then below you send me to a SUMO page. It may be more useful to send users right to the Firefox download page here. :unsolicted_advice:

views/unsupported.handlebars
```diff
@@ -1,11 +1,19 @@
...
+    <a id="update-firefox" href="https://support.mozilla.org/az/kb/update-firefox-latest-version">
```
pdehaan: If I'm reading this correctly, we're sending everybody to the Azerbaijani (az) locale of the page. We may just want to neuter the `/az/` portion of the URL.

public/locales/en-US/send.ftl
flodolo: Looks good to me.

Feel free to use the testpilot solution to ping automatically ;-)
https://github.com/mozilla/testpilot/blob/master/CODEOWNERS

views/unsupported.handlebars
ericawright: but the users already have firefox...they just need to know how to update it

views/unsupported.handlebars
flodolo: This should definitely be https://support.mozilla.org/kb/update-firefox-latest-version

public/locales/en-US/send.ftl
pdehaan: Cool, I just learnt something new today!
Filed as #343, which I can create a PR for shortly.

frontend/src/common.js
```diff
@@ -3,8 +3,31 @@ window.Raven.config(window.dsn).install();
...
+  sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
+    cd6: new Error('Firefox is outdated.')
+  })
+  location.replace('/unsupported/outdated');
+}
+
+if (navigator.userAgent.toLowerCase().indexOf('msie') > -1 ||
+    navigator.userAgent.toLowerCase().match(/trident\/7\./)) {
+  window.location.replace('/unsupported/ie');
+}
```
pdehaan: Before we go too far down a UA+regex abyss, not sure if we want to consider something like https://github.com/faisalman/ua-parser-js maybe.
https://www.npmjs.com/package/ua-parser-js#using-jqueryzepto-ua

frontend/src/common.js
```diff
@@ -3,8 +3,31 @@ window.Raven.config(window.dsn).install();
...
+    parseInt(navigator.userAgent.toLowerCase().match(/firefox\/*([^\n\r]*)\./)[1]) <= 49) {
+  sendEvent(isSender ? 'sender' : 'recipient', 'unsupported', {
+    cd6: new Error('Firefox is outdated.')
+  })
+  location.replace('/unsupported/outdated');
+}
+
+if (navigator.userAgent.toLowerCase().indexOf('msie') > -1 ||
+    navigator.userAgent.toLowerCase().match(/trident\/7\./)) {
+  window.location.replace('/unsupported/ie');
```
pdehaan: Do we need a `sendEvent()` here (before the `location.replace()`)?

SoftVision-CiprianMuresan:
@ericawright, using a local server I've run branch v49 on a Mac and connected remotely from my Windows machine. 
Unfortunately Internet Explorer does not show the "Browser not supported" page:

image


It does, however, work on Firefox 49 and lower.

image


And to answer your other question, in my opinion we should only block Microsoft Edge v.25.10586.672.0 . The other two versions of Edge work as expected (see Cosmin's comment in [#127](https://github.com/mozilla/send/issues/127#issuecomment-316321933)).

ericawright:
@SoftVision-CiprianMuresan @dannycoates @pdehaan (and others) we have some options to deal with IE. what's happening is that it doesn't support ES6. And it errors in our scripts then fails to run any part of them. So @dnarcese and I have come up with a few solutions. 
1. We create a separate script included in all html files that checks for IE, then redirects to our error page without any ES6 syntax.
2. We use babelify/browserify to interpret our code to es5 in the bundles. 

1 is easier, but we may end up never supporting IE if we go that route.
2 causes us to lose some of the benefits of new ES6 speed, though it shouldn't be too big a problem.

frontend/src/common.js
ericawright: good idea. @abhinadduri would you like a different event, or the same event as gets sent for outdated firefox?

frontend/src/common.js
ericawright: I'm not a huge fan of including npm modules for everything that seems like overkill. I'm open to it, but the regex is already written and really not a big problem, I don't see too many more UA checks being added in...

pdehaan:
I think +1 to using babel to transpile to ES5 for maximum compatibility (although we'd still need GCM support, so we wouldnt get that much compatibility from older browsers anyways).

Also, while trying to hack on UA options yesterday, I found https://github.com/biggora/express-useragent. Maybe we can just detect the IE browsers using Express middleware on the servers and do the redirect there for your option 1. Then it can easily use a custom layout if needed.

pdehaan:
@SoftVision-CiprianMuresan Any chance you can send us your IE user agent string: http://www.whatismyuseragent.net/?

ericawright:
@pdehaan turns out @dnarcese has ie11 set up so we've been using that this morning. 
"Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; .NET4.0C; .NET4.0E; rv:11.0) like Gecko”

pdehaan:
> And to answer your other question, in my opinion we should only block Microsoft Edge v.25.10586.672.0 .

@SoftVision-CiprianMuresan I guess of you have a UserAgent string for that too, it would be helpful.

SoftVision-CiprianMuresan:
Here's my Edge user-agent: 
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2486.0 Safari/537.36 Edge/13.10586

pdehaan:
Well, the IE one would fail, because the RegExp works for `/trident\/7\./i`, but neither of your IE user agent strings contain the "msie" substring.

```diff
+if (navigator.userAgent.toLowerCase().indexOf('msie') > -1 ||
+    !!navigator.userAgent.toLowerCase().match(/trident\/7\./)) {
+  window.location.replace('/unsupported/ie');
+}
```

ericawright:
@pdehaan this is true. I'll find another way to identify IE (outside of IE11)

ericawright:
we're struggling with babelify. Since @dannycoates talked about doing a new release today, maybe we merge this without the IE changes. Then either get babelify working or do a last minute fix using option one from above.
(the problem is that it's not correctly changing the `class` keyword if anyone wants to look into it)

@pdehaan 
> I take that back... re-reading your if(){}, it has an || not an &&, so the /trident/ regexp should still have been truthy-enough to make that work.

Yes, the 'msie' check is for IE < 11, and the 'trident' check is for IE11

pdehaan:
> we're struggling with babelify

I was starting to look into replacing browserify with webpack the other day, but then got side tracked by other projects... If I have a few minutes today I may see if I can get a basic webpack config+minification working, then try adding in babel into the stack-of-doom.

pdehaan:
But I'm 👍 for merging whatever we have, and filing new bugs for sorting out the older IE issues in a future bug/PR.

I really thought IE would have been caught by the GCM check anyways.  But maybe that's an issue w/ the promise being async and it trying to run the UA regexes before the GCM check has had a chance to fail out. Or something else entirely.

ericawright:
It seems that because of the ES6 syntax, IE just gives up and doesn't run anything if there is any ES6 syntax anywhere in the file. So the GCM check might have otherwise caught on it. it's the same with the redirect I'm trying to do. the code is correct. and will run in the console, but it sees the file as being corrupted.

ericawright:
Ok, I took out the IE code. So it's ready for review from whoever would like to give one :)

pdehaan:
Submitted new bug to re-add IE detection code: _#351 "Redirect IE users to an /unsupported/ie page"_.

PR #340, opened 2017-07-27
Remove duplicate entities in localized FTL files

flodolo:
Removing trailing spaces is irrelevant in this context (they should have one)


flodolo:
I see unrelated linter fails there were fixed in a later commit, merging without doing a rebase and another run.

PR #352, opened 2017-07-28
[WIP] add separate js file for checking ie

dnarcese:
Right now the localization doesn't work on the IE unsupported page. 
@dannycoates can you get localization running without ES6?


dannycoates:
lets put this check at the top of `jsconfig.handlebars` (maybe we need to change the `const`s to `var`s?)

dannycoates:
closing in favor or #353

PR #353, opened 2017-07-28
redirect ie to /unsupported

### LINKED ISSUE #351: Redirect IE users to an /unsupported/ie page
I think we backed out the IE and IE11 checks from #339, per @ericawright's https://github.com/mozilla/send/pull/339#issuecomment-318726696.
We should try and sort out the regexes and checks and add that back at some point.

dannycoates:
fixes #351

### DIFF
--- a/views/jsconfig.handlebars
+++ b/views/jsconfig.handlebars
@@ -1,8 +1,15 @@
+{{!-- This file should be es5 only --}}
+var isIE = /trident\/7\.|msie/i.test(navigator.userAgent);
+var isUnsupportedPage = /\/unsupported/.test(location.pathname);
+if (isIE && !isUnsupportedPage) {
+  window.location.replace('/unsupported/ie');
+}
+
 {{#if dsn}}
 window.dsn = '{{{dsn}}}';
 {{/if}}
 {{#if trackerId}}
 window.trackerId = '{{{trackerId}}}';
 {{/if}}
-const MAXFILESIZE = {{{maxFileSize}}};
-const EXPIRE_SECONDS = {{{expireSeconds}}};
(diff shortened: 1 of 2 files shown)


views/jsconfig.handlebars
```diff
@@ -1,8 +1,16 @@
+{{!-- This file should be es5 only --}}
+var isIE = navigator.userAgent.toLowerCase().indexOf('msie') > -1 ||
+    !!navigator.userAgent.toLowerCase().match(/trident\/7\./);
```
pdehaan: Instead of a double negation on `.match()`, we could possibly just use `.test()`:

```js
/trident\/7\./i.test(navigator.userAgent);
```

views/jsconfig.handlebars
```diff
@@ -1,8 +1,16 @@
...
+var isUnsupportedPage = location.pathname.includes('/unsupported');
```
pdehaan: I don't think we can use `.includes()` here, per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility

May have to just use a cheap `.indexOf('/unsupported') !== -1)` or similar.

dannycoates: this is what happens when I start a PR at 4:30 on Friday

ericawright:
So the redirect works, in much the same way that #352 did, but it still has the same issue in that the page shows up like this due to l10n not firing. (As well as some CSS issues, but that should hopefully be easily fixable once we have some text)
screen shot 2017-07-31 at 9 59 51 am

ericawright:
screen shot 2017-07-31 at 11 54 33 am
@dannycoates the console errors for you

PR #356, opened 2017-07-29
Get rid of console.log statements

### LINKED ISSUE #283: Get rid of console.log statements
Current status:

```sh
$ npm run lint

...

> firefox-send@0.2.0 lint:js /home/ubuntu/send
> eslint .


/home/ubuntu/send/frontend/src/download.js
  105:9  warning  Unexpected console statement  no-console
  156:9  warning  Unexpected console statement  no-console

/home/ubuntu/send/frontend/src/fileSender.js
  29:11  warning  Unexpected console statement  no-console
  31:11  warning  Unexpected console statement  no-console
  57:11  warning  Unexpected console statement  no-console

/home/ubuntu/send/frontend/src/upload.js
  212:9   warning  Unexpected console statement  no-console
  224:9   warning  Unexpected console statement  no-console
  300:11  warning  Unexpected console statement  no-console

✖ 8 problems (0 errors, 8 warnings)
```

We should either remove all unnecessary `console.log()` statements (and/or ESLint ignore any valid ones), or possibly just remove the ESLint rule to avoid console vomit in Circle-CI and when linting locally.

pdehaan:
Fixes #283

### DIFF
--- a/frontend/src/fileSender.js
+++ b/frontend/src/fileSender.js
@@ -46,10 +46,6 @@ class FileSender extends EventEmitter {
           },
           true,
           ['encrypt', 'decrypt']
-        )
-        .catch(err =>
-          // eslint-disable-next-line no-console
-          console.error('There was an error generating a crypto key')
         ),
       new Promise((resolve, reject) => {
         const reader = new FileReader();
--- a/frontend/src/fileSender.js
+++ b/frontend/src/fileSender.js
@@ -24,12 +24,6 @@ class FileSender extends EventEmitter {
         if (xhr.readyState === XMLHttpRequest.DONE) {
           resolve();
         }
-
-        if (xhr.status === 200) {
-          console.log('The file was successfully deleted.');
-        } else {
-          console.log('The file has expired, or has already been deleted.');
-        }
       };
 
       xhr.send(JSON.stringify({ delete_token: token }));
@@ -54,7 +48,8 @@ class FileSender extends EventEmitter {
           ['encrypt', 'decrypt']
         )
         .catch(err =>
-          console.log('There was an error generating a crypto key')
+          // eslint-disable-next-line no-console
+          console.error('There was an error generating a crypto key')
         ),
       new Promise((resolve, reject) => {
         const reader = new FileReader();
--- a/server/server.js
+++ b/server/server.js
@@ -33,12 +33,7 @@ function allLangs() {
 }
 
 function prodLangs() {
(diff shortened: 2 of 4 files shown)


frontend/src/fileSender.js
```diff
@@ -54,7 +48,8 @@ class FileSender extends EventEmitter {
           ['encrypt', 'decrypt']
         )
         .catch(err =>
```
dannycoates: we should just remove this catch

frontend/src/fileSender.js
```diff
@@ -24,12 +24,6 @@ class FileSender extends EventEmitter {
         if (xhr.readyState === XMLHttpRequest.DONE) {
           resolve();
         }
-
-        if (xhr.status === 200) {
-          console.log('The file was successfully deleted.');
-        } else {
-          console.log('The file has expired, or has already been deleted.');
-        }
```
pdehaan: This felt weird to kill, but wasn't really "adding" anything to the user experience.
Not sure if we should be doing some metrics logging here instead of just vanilla `console.log()`.

frontend/src/fileSender.js
```diff
@@ -54,7 +48,8 @@ class FileSender extends EventEmitter {
           ['encrypt', 'decrypt']
         )
         .catch(err =>
-          console.log('There was an error generating a crypto key')
+          // eslint-disable-next-line no-console
+          console.error('There was an error generating a crypto key')
```
pdehaan: Kept this since removing the `catch()` would alter the code paths. This `catch()` is actually swallowing the error, IIUC. So if `window.crypto.subtle.generateKey()` fails for whatever reason, we're intercepting the error and just ignoring it, allowing `Promise.all()` to resolve with potentially an `undefined` or `null` `secretKey`. 

Or not, maybe I'm wrong. 🤷‍♀️

server/server.js
```diff
@@ -33,6 +33,7 @@ function allLangs() {
 }
 
 function prodLangs() {
+  // eslint-disable-next-line security/detect-non-literal-require
```
pdehaan: Unrelated, but this was the only remaining ESLint warning.
Looking again, that could have been normalized out by potentially doing `require('../package.json')` and letting Node.js resolve paths internally (like it does elsewhere for `require('./utils')`.

dannycoates: yeah, why the f did i do it like this? 🍌

PR #363, opened 2017-07-31
fix ie css

ericawright:
Fixes the footer and logo css on the unsupported page.
@dannycoates 

something that wasn't addressed in the last PR, but was brought up briefly - did we want to add a different GA event depending on the reason a page is not supported, like for IE?


dannycoates:
> did we want to add a different GA event depending on the reason a page is not supported, like for IE?

I'll file an issue

PR #365, opened 2017-07-31
revert the IE fix to fix footer on chrome

dannycoates:
The previous commit was to fix an IE issue, but it caused the footer to float over content on Chrome.

This worked for me, but maybe there's a better way?

### DIFF
--- a/public/main.css
+++ b/public/main.css
@@ -21,7 +21,6 @@ body {
   flex-direction: column;
   margin: 0;
   min-height: 100vh;
-  height: 100%;
   position: relative;
 }
 


ericawright:
on chrome does min-height 100% work, instead of VH?

dannycoates:
> on chrome does min-height 100% work, instead of VH?

min-height: 100% **without** a height: 100% seems to work. is that better?

ericawright:
I'll try on ie, one min

ericawright:
nope, no good. ie does not like, "min". I suppose we can add an attribute to the `body` when we redirect to `unsupported/ie` then change css to be 
```
body[ie] {
  height: 100%;
}
```

dannycoates:
how do you feel about this patch and releasing with the footer broken in IE?

ericawright:
sure, better ie is broken than chrome

PR #379, opened 2017-08-01
WIP: added web push notifications for download completion

abhinadduri:


public/serviceWorker.js
```diff
@@ -0,0 +1,7 @@
+self.addEventListener('push', function(event) {
+  event.waitUntil(
+    self.registration.showNotification('Firefox Send', {
+      body: event.data.text() + ' has been downloaded.',
```
pdehaan: Does this need to be localized?

frontend/src/upload.js
```diff
@@ -26,6 +26,34 @@ if (storage.has('referrer')) {
...
+      .then(function(subscription) {
+        if (subscription) {
+          return subscription;
+        }
+        return registration.pushManager.subscribe({ userVisibleOnly: true });
+      });
+    }).then(function(subscription) {
+      const rawKey = subscription.getKey ? subscription.getKey('p256dh') : '';
+      const key = rawKey ?
+        btoa(String.fromCharCode.apply(null, new Uint8Array(rawKey))) : '';
```
pdehaan: Only duplicated twice, but not sure if you want to wrap this in a helper function.

server/server.js
```diff
@@ -177,6 +178,17 @@ app.get('/assets/download/:id', (req, res) => {
...
+            webPush.sendNotification({
+              endpoint: meta.endpoint,
+              keys: {
+                p256dh: meta.key,
+                auth: meta.authSecret
+              }
+            },
+            meta.filename)
+            .catch(err => {
+              console.log(err);
```
pdehaan: Do we need to log this error somewhere? Or is it a non-important error?

abhinadduri: It's a non-important error. I actually meant to use `.catch(err=>0)`, but I put it there for debugging purposes for now.

public/serviceWorker.js
abhinadduri: Yeah, nice catch. We'll have to wrap this in a promise reaching out to l20n.

PR #386, opened 2017-08-02
fix percentage view on mobile layout

ariestiyansyah:
Percentage number doesn't fit in the middle when using mobile browser, chrome, firefox and safari on desktop, this commit is simple solution.

**Before**

**After**

### DIFF
--- a/public/main.css
+++ b/public/main.css
@@ -389,7 +389,6 @@ tbody {
   letter-spacing: -0.78px;
   font-family: 'Segoe UI', 'SF Pro Text', sans-serif;
   top: 53px;
-  left: 246.75px;
   width: 98.5px;
   -moz-user-select: none;
   -ms-user-select: none;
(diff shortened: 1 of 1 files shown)


ericawright:
arguably, it is more important to support mobile than older FF versions if we can't find a fix for this easily.

ariestiyansyah:
For me mobile first is important and also considering user who use different browser and screen size is good thing.

ericawright:
I checked, and it will work for 51 and 52 if you add 
```
.percentage {
...
  left: 50%;
  transform: translateX(-50%);
```
Plus your removal of the left rule. While you're there, could you remove the `width: 98.5px;` rule? As I don't think it does anything.
I'd add it myself to your PR but I can't seem to figure out if I can do that or not ;)

Thanks for this!

ariestiyansyah:
@ericawright thanks Erica, commit updated 5576bc065230739a721c65728e2c056cd0bf0118

ericawright:
(the test timing out is not your fault) Merging, thank you!

Issue #382, opened 2017-08-02
Odd behavior when hovering over Send icon/wordmark in header


pdehaan:
Found in https://send.dev.mozaws.net/__version__ (bca7948):

```js
{
  commit: "bca7948",
  source: "https://github.com/mozilla/send/",
  version: "v1.0.2"
}
```

---

### Steps to reproduce:
1. Go to https://send.dev.mozaws.net.
1. Hover over the Send logo/wordmark in header.

### Actual results:
Even though you're hovered over the logo, and the target HREF is https://send.dev.mozaws.net/, the "Firefox Test Pilot" text highlights.

<img width="328" alt="firefox_send" src="https://user-images.githubusercontent.com/557895/28858094-1538b620-7703-11e7-85e4-99845f2cadd9.png">

### Expected results:
The "Test Pilot" text shouldn't highlight unless you're hovering over the https://testpilot.firefox.com/ link. Not sure if hovering over the "Send" logo or wordmark should cause "Send" to turn blue.

wresuolc:
"This was legit at first" --erica

PR #392, opened 2017-08-02
Adjust hover behavior on send-logo (#382)

weihanglo:
I adjusted highlight targets to `h1` and `a`. Now it looks like this:

untitled

Not quite sure if this fits the original design.

### DIFF
--- a/public/main.css
+++ b/public/main.css
@@ -39,6 +39,11 @@ body {
   align-items: center;
 }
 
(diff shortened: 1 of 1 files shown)


weihanglo:
Should I amend the previous commit, or two separate commits is OK?

ericawright:
@weihanglo whichever you prefer is fine

Issue #359, opened 2017-07-30
Transpile front end code into ES5?


pdehaan:
I was playing around trying to replace browserify with webpack+babel, and failing quite spectacurly.
I think the latest error I was getting was re: `class` keyword in an imported ["testpilot-ga" module](https://github.com/mozilla/testpilot-ga/blob/master/.babelrc).

Looks like testpilot-ga publishes a babel version itself (yay) but targets a baseline env of "Firefox 54", whereas we're targeting older browsers. Not sure if that means we'd have to re-babel that dependency and set our `env.targets.browsers` to something like "> 2%,  ie 8, safari > 9".

dannycoates:
done

PR #396, opened 2017-08-02
add babel-polyfill

### LINKED ISSUE #395: Object.entries is not a function
https://sentry.prod.mozaws.net/operations/send-prod/issues/632248/

dannycoates:
fixes #395

and before @pdehaan mentions webpack, i know, but not right now

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "redis": "^2.7.1"
   },
   "devDependencies": {
+    "babel-polyfill": "^6.23.0",
     "browserify": "^14.4.0",
     "eslint": "^4.3.0",
     "eslint-plugin-mocha": "^4.11.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -79,7 +80,7 @@
     "build:upload": "browserify frontend/src/upload.js -g uglifyify -o public/upload.js",
     "build:download": "browserify frontend/src/download.js -g uglifyify -o public/download.js",
     "build:version": "node scripts/version",
-    "build:l10n": "cp node_modules/l20n/dist/web/l20n.min.js public",
+    "build:3rd": "cp node_modules/l20n/dist/web/l20n.min.js public && cp node_modules/babel-polyfill/dist/polyfill.min.js public",
```
pdehaan: I'll give you a 💵 if we rename this to something more grep-able like "build:vendor".

You could even combine the two:

```sh
cp node_modules/l20n/dist/web/l20n.min.js node_modules/babel-polyfill/dist/polyfill.min.js public
```

pdehaan:
Do we know where the `Object.entries()` is coming from? I didn't see it in our source (via a quick `git grep`).

I mean, **babel-polyfill** seems like a big [96KB] hammer for an API call:

```sh
$ ls -lash node_modules/babel-polyfill/dist/polyfill.min.js | awk '{ print $6, $10 }'

96K node_modules/babel-polyfill/dist/polyfill.min.js
```

dannycoates:
> I mean, babel-polyfill seems like a big [96KB] hammer for an API call

it is, but this isn't the only error in sentry the polyfill will fix

pdehaan:
> it is, but this isn't the only error in sentry the polyfill will fix

👍 Great point! Ship it. I think we were considering doing webpack+babel+transpiling (#359) anyways, so this is a great start.

Issue #383, opened 2017-08-02
Upload an empty file returns incorrect message


yfdyh000:
Try upload an 0 bytes file to testing the functionality.

The Send page alert the `Uploading multiple files or a folder is currently not supported.` (`uploadPageMultipleFilesAlert`).

pdehaan:
Not sure exactly what the code is doing, but I believe this is coming from [/frontend/src/upload.js:143](https://github.com/mozilla/send/blob/f1c894d14fce7a514428914ffa69b82355508f42/frontend/src/upload.js#L143), where we explicitly check for a file size of 0:

```js
event.originalEvent.dataTransfer.files[0].size === 0
```

ericawright:
That check is in there in order to catch when the user uploads a folder. The "file" size comes through as 0 in that case as well.

ehuggett:
I can upload 0byte files on production and dev.

I did manage to grab this screenshot from dev
![send_nan](https://user-images.githubusercontent.com/8090731/29754963-d874219a-8b87-11e7-98cb-3d5cc94fb0c3.png)

234B sounds plausible for posting an "empty" encrypted file (probably 32Bytes itself?) as an attachment on a form that also has a few metadata fields and headers (filename, content-type etc), its the total size of the form (and its file)?

NaN is of course "not a number", presumably something has been divided by the file size?

I tried to upload the 0byte file to github as "proof" that it really was empty, but github won't let me 😄 

![screenshot](https://user-images.githubusercontent.com/8090731/29755005-5ba4346a-8b88-11e7-90e9-f530de1c9996.png)

himanish-star:
Can I work on this issue ?

dannycoates:
> Can I work on this issue ?

Empty files can no longer be uploaded so this was already fixed at some point.

PR #399, opened 2017-08-02
Possible fix for issue #383

gvanwaelvelde:
This could possibly fix issue #383 .
Just a suggestion...


ericawright:
I tested it with a single file, size 0bytes, and I still received the "Uploading multiple files or a folder is currently not supported." message

gvanwaelvelde:
Maybe we could use just one error message for both a folder and an empty file?
"Uploading an empty file or a folder is not allowed"
This is not ideal, and may even be confusing for some users, but when we can't differentiate between an empty file and a folder, we may have no choice.

pdehaan:
Wow, this is pretty interesting... I'm seeing some different behavior on Chrome/Firefox with regards to folder sizes and how gracefully we handle empty folders vs folders with 1 file versus multiple files.

---

I have a file called "zerobytes.txt" which is just an empty file on my desktop.
If I drag that file into Firefox Send (using Firefox), it correctly says "Uploading an empty file is not allowed":

firefox_send

Chrome seems to agree:

firefox_send

firefox_send

---

Attempting to upload an empty directory into Firefox gives the same error message of "Uploading an empty file is not allowed". Not sure how to avoid that bit, since it still appears as a single file in a filelist with 0 bytes:

firefox_send

This is where Chrome starts to get a bit unexpected. Dropping an empty folder onto the drop zone tries uploading a directory, but takes me to the "Something went wrong!" page with an error. But note that Chrome reports an empty directory as having a size of 68 bytes.

firefox_send

Anyways. Not sure if dropping folders is a common-ish use case, or if that explains an error rate for "upload-stopped" with a `cd2` of "errored".

---

Chrome behavior seems equally baffling when you try dragging a folder with exactly 1 file versus a folder with multiple files.

The only consistency I see with dropping folders, is that Chrome (Release) and Firefox (Beta) both return a `FileList[0].type = ""` when dropping folders. But that feels way too brittle to try writing logic around.

dannycoates:
I think other changes have superseded the need for this

PR #402, opened 2017-08-02
filter the hash from error reports

dannycoates:
we don't want to know it

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,6 @@
     "test": "npm-run-all test:*",
     "test:unit": "mocha test/unit",
     "test:server": "mocha test/server",
-    "test:browser": "browserify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d && node test/frontend/driver.js"
+    "test--browser": "browserify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d && node test/frontend/driver.js"
   }
 }
--- a/views/jsconfig.handlebars
+++ b/views/jsconfig.handlebars
@@ -5,11 +5,24 @@ if (isIE && !isUnsupportedPage) {
   window.location.replace('/unsupported/ie');
 }
 
-{{#if dsn}}
-window.dsn = '{{{dsn}}}';
+{{#if sentryId}}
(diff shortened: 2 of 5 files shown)


views/jsconfig.handlebars
```diff
@@ -5,11 +5,24 @@ if (isIE && !isUnsupportedPage) {
...
+{{#if sentryId}}
+var RAVEN_CONFIG = {
+  release: '{{{version}}}',
+  tags: {
+    commit: '{{{commit}}}'
+  },
+  dataCallback: function (data) {
+    var hash = window.location.hash;
+    if (hash) {
+      return JSON.parse(JSON.stringify(data).replace(new Regexp(hash.slice(1), 'g'), ''));
```
pdehaan: Should be `RegExp` (per [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/RegExp))? 

I was going to see if ESLint caught that, but I guess this is a weird hybrid .handlebars file, which linting would never see.

Not sure if we need to wrap any of that in a `try..catch()` in case something implodes unexpectedly.

package.json
```diff
@@ -92,6 +92,6 @@
     "test": "npm-run-all test:*",
     "test:unit": "mocha test/unit",
     "test:server": "mocha test/server",
-    "test:browser": "browserify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d && node test/frontend/driver.js"
+    "test--browser": "browserify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d && node test/frontend/driver.js"
```
pdehaan: 💯

dannycoates:
I generated a test error, and the fragment is no longer included https://sentry.prod.mozaws.net/operations/send-dev/issues/633106/

views/jsconfig.handlebars
dannycoates: I think throwing is better in this case. It has the least surprising result, the report isn't sent. Returning `null` sends the original `data` including fragment, and returning `{}` also doesn't send the report but is less obvious what it might do.

Issue #325, opened 2017-07-26
[style] use async/await


dannycoates:
It might be fun to use `async/await` syntax on the server.

Here's a sample snippet:

```js
app.get('/download/:id', async (req, res) => {
  const id = req.params.id;
  if (!validateID(id)) {
    return res.sendStatus(404);
  }
  try {
    const filename = await storage.filename(id);
    const contentLength = await storage.length(id);
    const timeToExpiry = await storage.ttl(id);
    res.render('download', {
      filename: decodeURIComponent(filename),
      filesize: bytes(contentLength),
      sizeInBytes: contentLength,
      timeToExpiry: timeToExpiry
    });
  } catch (e) {
    res.status(404).render('notfound');
  }
});
```

pdehaan:
💯 

Related, doing something like #64 to convert Redis calls to a promise style instead of callback style, would make it play much better w/ `async/await`.
As would switching to the already promise-based `fetch()` API, circa #32 (although that's more of a front end optimization).

weihanglo:
I changed all promises into async functions in express at weihanglo/send@bbe111a.

However, in **storage.js**, methods on Redis client seem cannot be promisified by NodeJS 8 `util.promisify`. I thought this issue may be related to 424eb2c. Any idea?

dannycoates:
@weihanglo thanks! Feel free to submit a PR and we can discuss the details there.

pdehaan:
Awesome work, @weihanglo!

There are a few more Promise notes buried in #64 (most notably the https://github.com/mozilla/send/issues/64#issuecomment-310733702 which has a pointer to the official Redis+Promise documentation on using [**bluebird**](http://npm.im/bluebird) to promisify the entire RedisClient using something like:

```js
bluebird.promisifyAll(redis.RedisClient.prototype);
```

I only briefly played with their sample code, but I'm sure we could figure it out if you need a second pair of eyes.

gautamkrishnar:
Can this issue be closed? I think it's fixed now...

pdehaan:
@gautamkrishnar I think we're still keeping this open until we try and promisify the server/storage.js bits.

Issue #64, opened 2017-06-08
Replace callbacks w/ promise wrappers with util.promisify?


pdehaan:
In the spirit of needless optimizations, we may be able to reduce the code by a few bytes if we use the new Node 8 [`util.promisify()`](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) in a couple places.

Ref: http://2ality.com/2017/05/util-promisify.html

For example, instead of [this](https://github.com/mozilla/something-awesome/blob/24abf669b02a6bffee5318c76a6301453cefde22/server/storage.js#L44-L54):

```js
function filename(id) {
  return new Promise((resolve, reject) => {
    redis_client.hget(id, 'filename', (err, reply) => {
      if (!err) {
        resolve(reply);
      } else {
        reject();
      }
    });
  });
}
```

... we _may_ be able to shorten to:

```js
const { promisify } = require('util');
// ...

function filename(id) {
  const hget = promisify(redis_client.hget);
  return hget(id, 'filename');
}
```

And in theory the above will resolve with the `reply` object or reject with the `err` object.
I don't think we can use it in all the places we're using wrapper Promise objects though.

pdehaan:
Another example:

```js
function localDelete(id, delete_token) {
  return new Promise((resolve, reject) => {
    redis_client.hget(id, 'delete', (err, reply) => {
      if (!reply || delete_token !== reply) {
        reject();
      } else {
        redis_client.del(id);
        resolve(fs.unlinkSync(__dirname + '/../static/' + id));
      }
    });
  });
}

function localForceDelete(id) {
  return new Promise((resolve, reject) => {
    redis_client.del(id);
    resolve(fs.unlinkSync(__dirname + '/../static/' + id));
  });
}
```

vs

```js
function localDelete(id, delete_token) {
  const hget = promisify(redis_client.hget);
  return hget(id, 'delete')
    .then((reply) => {
      if (!reply || reply !== delete_token) {
        throw new Error(`Uh oh, unable to delete ${id}: ${reply}`);
      }
      return localForceDelete(id);
    });
}

function localForceDelete(id) {
  redis_client.del(id);
  return Promise.resolve(fs.unlinkSync(path.join(__dirname, '..', 'static', id)));
}
```

pdehaan:
Potentially better approach (at the expense of yet another dependency): https://github.com/NodeRedis/node_redis#promises

PR #404, opened 2017-08-03
Use async function instead of promise (#325)

weihanglo:
Changed all promises into async functions for express.

I have attempted to promsify all Redis client methods as #64 described. It failed but I have no idea why, and I cannot log directly in [storage.js](https://github.com/mozilla/send/blob/424eb2c37ad50ab7778985edbea029e8683e4720/server/storage.js). Any help?

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -205,7 +205,7 @@ app.post('/delete/:id', async (req, res) => {
   }
 
   try {
-   const err = await storage.delete(id, delete_token);
+    const err = await storage.delete(id, delete_token);
     if (!err) {
       log.info('Deleted:', id);
       res.sendStatus(200);
--- a/server/server.js
+++ b/server/server.js
@@ -117,93 +117,79 @@ app.get('/jsconfig.js', (req, res) => {
   });
 });
 
-app.get('/exists/:id', (req, res) => {
+app.get('/exists/:id', async (req, res) => {
   const id = req.params.id;
   if (!validateID(id)) {
     res.sendStatus(404);
     return;
   }
 
-  storage
-    .exists(id)
-    .then(() => {
-      res.sendStatus(200);
-    })
-    .catch(err => res.sendStatus(404));
+  try {
+    await storage.exists(id);
+    res.sendStatus(200);
+  } catch (e) {
+    res.sendStatus(404);
+  }
 });
 
-app.get('/download/:id', (req, res) => {
+app.get('/download/:id', async (req, res) => {
   const id = req.params.id;
   if (!validateID(id)) {
     res.sendStatus(404);
     return;
   }
 
-  storage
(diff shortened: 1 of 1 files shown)


server/server.js
```diff
@@ -117,93 +117,79 @@ app.get('/jsconfig.js', (req, res) => {
...
-          });
-        });
-      });
-    })
-    .catch(() => {
-      res.status(404).render('notfound');
+  try {
+    const filename = await storage.filename(id);
+    const contentLength = await storage.length(id);
+    const timeToExpiry = storage.ttl(id);
```
dannycoates: this needs an `await`

weihanglo: Done 596ad87!

dannycoates:
As you've discovered converting storage.js is going to be more complicated. Let's do that in a separate PR. The easiest strategy is probably to add bluebird, as recommended in the node_redis [readme](https://github.com/NodeRedis/node_redis#promises). We'll also probably need to adjust the code a bit if that changes the behavior or rejecting, i.e. change some try/catches into checking result values.

dannycoates:
Thanks @weihanglo! I'll verify this and merge it in the morning :)

server/server.js
```diff
@@ -218,15 +204,15 @@ app.post('/delete/:id', (req, res) => {
...
-    .delete(id, delete_token)
-    .then(err => {
-      if (!err) {
-        log.info('Deleted:', id);
-        res.sendStatus(200);
-      }
-    })
-    .catch(err => res.sendStatus(404));
+  try {
+   const err = await storage.delete(id, delete_token);
```
pdehaan: Super minor nit, but I think the indenting is slightly off here. Not sure if you want to try risking running `npm run format` and running the **prettier** formatter against the codebase.

server/server.js
```diff
@@ -254,48 +240,50 @@ app.post('/upload', (req, res, next) => {
...
-      .forceDelete(newId)
-      .then(err => {
-        if (!err) {
-          log.info('Deleted:', newId);
-        }
-      })
-      .catch(err => {
-        log.info('DeleteError:', newId);
-      });
+  req.on('close', async err => {
```
pdehaan: Unrelated to PR, but I just noticed that the `.on('close')` may return an `err` response here (apparently), which we aren't checking, but then we're overwriting the `err` value below on line 266.  🤷‍♀️ 
Behavior looks the same as it was previously, but there may be dragons here. Very small, uninteresting dragons.

server/server.js
weihanglo: Sorry I totally forgot it! Maybe we should add pre-commit or pre-push hook to prevent these kind of mistake.

dannycoates:
Thanks again, @weihanglo!

PR #406, opened 2017-08-03
Add noscript tag

### LINKED ISSUE #373: Add noscript tag?
Not even sure what the world does about `<noscript>` anymore since JavaScript Rules Everything Around Me.

But our site looks slightly suboptimal (since we rely heavily on JavaScript+l20n).

firefox_send

Not sure how to properly fix it. I'm guessing we add `hidden="hidden"` to `<div id="page-one">` as well, and then remove that using jQuery, then we can add a `<noscript>` block which briefly says  something like "Firefox Send relies on JavaScript in order to encrypt and transfer files. Please enable JavaScript and try again." (with an optional link to an FAQ answer about why JavaScript is awesome, and how we need it in order to safely encrypt/decrypt the file on the user's computer).

As for the missing footer links, I think we have to do the same /unsupported/ie fallback we did where we just hardcode some en-US strings into the footer so they at least have some legal links if they cannot load l20n for some reason. In the case of the local /legal page, that may be a small comfort unless we hardcode all the en-US strings there too:

firefox_send

pdehaan:
You can test this locally by toggling the `javascript.enabled` Boolean in _about:config_.

With a locale set of "fr", this totally looks like:

firefox_send

Fixes #373

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -19,10 +19,12 @@ particular).
 
 ## Why does Firefox Send require JavaScript?
 
-- We use JavaScript to encrypt and decrypt files locally on the client instead of the server.
-- We use JavaScript to render the user interface of Send.
-- We use JavaScript to manage translate the website into [various different languages](https://github.com/mozilla/send#localization).
-- We use JavaScript to collect data to help us improve Send in accordance with our [Terms &amp; Privacy](https://send.firefox.com/legal).
+Firefox Send uses JavaScript to:
+
+- Encrypt and decrypt files locally on the client instead of the server.
+- Render the user interface.
+- Manage translations on the website into [various different languages](https://github.com/mozilla/send#localization).
+- Collect data to help us improve Send in accordance with our [Terms & Privacy](https://send.firefox.com/legal).
 
 Since Send is an open source project, you can see all of the cool ways we use JavaScript by [examining our code](https://github.com/mozilla/send/).
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -17,6 +17,15 @@ Many browsers support this standard and should work fine, but some have not
 implemented it yet (mobile browsers lag behind on this, in
 particular).
 
+## Why does Firefox Send require JavaScript?
+
+- We use JavaScript to encrypt and decrypt files locally on the client instead of the server.
+- We use JavaScript to render the user interface of Send.
+- We use JavaScript to manage translate the website into [various different languages](https://github.com/mozilla/send#localization).
+- We use JavaScript to collect data to help us improve Send in accordance with our [Terms &amp; Privacy](https://send.firefox.com/legal).
+
+Since Send is an open source project, you can see all of the cool ways we use JavaScript by [examining our code](https://github.com/mozilla/send/).
+
 ## How long are files available for?
 
 Files are available to be downloaded for 24 hours, after which they are removed
--- a/views/index.handlebars
+++ b/views/index.handlebars
@@ -1,4 +1,4 @@
-<div id="page-one">
+<div id="page-one" hidden>
   <script src="/upload.js"></script>
   <div class="title" data-l10n-id="uploadPageHeader"></div>
   <div class="description">
--- a/views/layouts/main.handlebars
+++ b/views/layouts/main.handlebars
@@ -39,11 +39,11 @@
   <div class="footer">
     <div class="legal-links">
       <a href="https://www.mozilla.org"><img class="mozilla-logo" src="/resources/mozilla-logo.svg"/></a>
-      <a href="https://www.mozilla.org/about/legal" data-l10n-id="footerLinkLegal"></a>
-      <a href="https://testpilot.firefox.com/about" data-l10n-id="footerLinkAbout"></a>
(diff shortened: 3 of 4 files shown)


docs/faq.md
```diff
@@ -17,6 +17,15 @@ Many browsers support this standard and should work fine, but some have not
...
 particular).
 
+## Why does Firefox Send require JavaScript?
+
+- We use JavaScript to encrypt and decrypt files locally on the client instead of the server.
+- We use JavaScript to render the user interface of Send.
+- We use JavaScript to manage translate the website into [various different languages](https://github.com/mozilla/send#localization).
+- We use JavaScript to collect data to help us improve Send in accordance with our [Terms &amp; Privacy](https://send.firefox.com/legal).
+
+Since Send is an open source project, you can see all of the cool ways we use JavaScript by [examining our code](https://github.com/mozilla/send/).
```
pdehaan: Copied the general concept/text from [**mozilla/testpilot** docs/faq.md](https://github.com/mozilla/testpilot/blob/master/docs/faq.md#why-does-test-pilot-require-javascript), but could probably use a bit of tweaking.

views/index.handlebars
```diff
@@ -1,4 +1,4 @@
-<div id="page-one">
+<div id="page-one" hidden>
```
pdehaan: Hides the content by default on the homepage, and then we show the content on `document.ready` using jQuery in frontend/src/upload.js above (assuming the user has JavaScript enabled).

views/layouts/main.handlebars
```diff
@@ -29,16 +29,21 @@
     <a href="https://qsurvey.mozilla.com/s3/txp-firefox-send" rel="noreferrer noopener" class="feedback" target="_blank" data-l10n-id="siteFeedback">Feedback</a>
   </header>
   <div class="all">
+    <noscript>
+      <h2>Firefox Send requires JavaScript</h2>
+      <p><a href="https://github.com/mozilla/send/blob/master/docs/faq.md#why-does-firefox-send-require-javascript" target="_blank" rel="noreferrer noopener">Why does Firefox Send require JavaScript?</a></p>
+      <p>Please enable JavaScript and try again.</p>
+    </noscript>
```
pdehaan: This is all en-US only. Sorry. But since l20n doesn't work without JavaScript, we're a bit stuck unless we do weird workarounds like some meta refresh to localized copies of a page.

views/layouts/main.handlebars
```diff
@@ -29,16 +29,21 @@
...
-      <a href="https://www.mozilla.org/about/legal" data-l10n-id="footerLinkLegal"></a>
-      <a href="https://testpilot.firefox.com/about" data-l10n-id="footerLinkAbout"></a>
-      <a href="/legal" data-l10n-id="footerLinkPrivacy"></a>
-      <a href="/legal" data-l10n-id="footerLinkTerms"></a>
-      <a href="https://www.mozilla.org/privacy/websites/#cookies" data-l10n-id="footerLinkCookies"></a>
+      <a href="https://www.mozilla.org/about/legal" data-l10n-id="footerLinkLegal">Legal</a>
+      <a href="https://testpilot.firefox.com/about" data-l10n-id="footerLinkAbout">About Test Pilot</a>
+      <a href="/legal" data-l10n-id="footerLinkPrivacy">Privacy</a>
+      <a href="/legal" data-l10n-id="footerLinkTerms">Terms</a>
+      <a href="https://www.mozilla.org/privacy/websites/#cookies" data-l10n-id="footerLinkCookies">Cookies</a>
```
pdehaan: Again, all the defaults for the footer link text are in en-US, but will still localize correctly w/ l20n if they have JavaScript enabled.

pdehaan:
@dannycoates @flodolo Any last minute objections before I press the big, green button?

views/layouts/main.handlebars
flodolo: Ugh :-\

@stasm can we try to think if it's possible to provide a "solution", or at least a suggested workaround, for this?

views/layouts/main.handlebars
flodolo: @stasm Do we risk to get into trouble? I know we had in the past, with l20n and react fighting over updating this content, but we weren't using fluent-react

docs/faq.md
flodolo: I would really avoid repeating "We use JavaScript".
```
We use JavaScript to:
* Encrypt …
* Render …
etc.
```

I also doubt you need to encode the ampersand in a markdown file.

views/layouts/main.handlebars
stasm: Unless there are a11y concerns, I'm quite OK with this here being the suggested solution. Can we assume that people who disable JavaScript are tech-savvy enough to understand this message?

views/layouts/main.handlebars
stasm: There doesn't seem to be any client-side framework managing the rendering of these elements. If I'm not mistaken, it's a static template rendered by the server. L20n will remove the child elements and replace them with translations.

pdehaan: Yeah, we aren't using React/fluent-react on this project, just vanilla jQuery and l20n.

PR #409, opened 2017-08-03
Handle copy clipboard disabled

### LINKED ISSUE #108: Service doesn't notice that copy to clipboard is disabled and that the copy failed
Steps:

1. go into `about:config`
2. set `dom.allow_cut_copy` to `false`

Expected:

The "copy" button turns to "failed." or is grayed out.

Actual:

The "copy" button changes to an enthusiastic (but wrong) "copied!".

Johann-S:
Should handle disabled copy to clipboard.

But I have to find a way for something better than that : 
image

Fixes #108

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -18,6 +18,21 @@ if (storage.has('referrer')) {
   window.referrer = 'external';
 }
 
+const allowedCopy = () => {
+  const support = !!document.queryCommandSupported;
+  return support ? document.queryCommandSupported('copy') : false;
+}
+
+const copyToClipboard = (value) => {
+  const aux = document.createElement('input');
+  aux.setAttribute('value', value);
+  document.body.appendChild(aux);
+  aux.select();
+  const result = document.execCommand('copy');
+  document.body.removeChild(aux);
+  return result;
+}
+
 $(document).ready(function() {
   $('#file-upload').change(onUpload);
 
@@ -57,7 +72,7 @@ $(document).ready(function() {
   $('body').on('dragover', allowDrop).on('drop', onUpload);
   // reset copy button
   const $copyBtn = $('#copy-btn');
-  $copyBtn.attr('disabled', false);
+  $copyBtn.attr('disabled', !allowedCopy());
   $('#link').attr('disabled', false);
   $copyBtn.attr('data-l10n-id', 'copyUrlFormButton');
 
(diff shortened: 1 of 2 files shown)


frontend/src/upload.js
```diff
@@ -18,6 +18,21 @@ if (storage.has('referrer')) {
   window.referrer = 'external';
 }
 
+const allowedCopy = () => {
+  const support = !!document.queryCommandSupported
+  return support ? document.queryCommandSupported('copy') : false
```
Johann-S: Oups I should add semicolons here

frontend/src/upload.js
```diff
@@ -79,23 +94,21 @@ $(document).ready(function() {
...
-    $('#link').attr('disabled', true);
-    $copyBtn.html(
-      '<img src="/resources/check-16.svg" class="icon-check"></img>'
-    );
-    window.setTimeout(() => {
-      $copyBtn.attr('disabled', false);
-      $('#link').attr('disabled', false);
-      $copyBtn.attr('data-l10n-id', 'copyUrlFormButton');
-    }, 3000);
+    const copyResult = copyToClipboard($('#link').attr('value'));
```
Johann-S: `copyResult` not necessary

ericawright: 👍

public/main.css
```diff
@@ -504,6 +509,7 @@ tbody {
   background: #05a700;
   border: 1px solid #05a700;
   cursor: auto;
+  opacity: 0.3;
```
ericawright: what's this for?

frontend/src/upload.js
```diff
@@ -345,6 +358,9 @@ $(document).ready(function() {
       class: 'icon-copy',
       'data-l10n-id': 'copyUrlHover'
     });
+    if (!allowedCopy()) {
+      $copyIcon.addClass('disabled');
```
ericawright: since 'disabled' is a really common, can we add a 'disabled' attribute instead? Then check the css with [disabled]
we already have a rule for `#copy-btn:disabled` that can be reused.

frontend/src/upload.js
```diff
@@ -79,23 +94,21 @@ $(document).ready(function() {
     sendEvent('sender', 'copied', {
```
Johann-S: Sure 👍

public/main.css
Johann-S: Just a question of opinion but for me it's a better visual feedbacks for the users

frontend/src/upload.js
Johann-S: Sure 👍

public/main.css
ericawright: oh sorry, I read it wrong. I thought it was within `.popup`. this is good

frontend/src/upload.js
Johann-S: Sure that's what I'll do 👍

ericawright:
thanks so much @Johann-S for taking this one :)

frontend/src/upload.js
Johann-S: I'll push changes you asked, but for this change `#copy-btn:disabled` rule cannot be reused because `.icon-copy` is an image but I used `disabled` attribute instead of `.disabled`

Johann-S:
You're welcome @ericawright a pleasure to help Mozilla 👍 

Just a question out of the scope of this PR, are you open for some refactorisation of your front JavaScript because I noticed some bad practices (events alias, duplicate selectors...) ?

ericawright:
@Johann-S of course, you can open a bug explaining each of the changes needed, or you can just go ahead and open a PR.

Johann-S:
Oh ok so should I revert this change : https://github.com/Johann-S/send/blob/patch-1/frontend/src/upload.js#L113-L119 ?

ericawright:
It's a good change, but it's unrelated to the current changes. if it's just that then it's fine, might as well leave it in. but if there are other unrelated changes then a separate PR would make it clearer to review

Johann-S:
No it's the only one 👍 

My last concern about this PR, it's when we want to copy an old uploaded file link and if copy/paste command isn't available currently it's isn't possible 🤔 

That's why I shown this image : 
image

Should I change the icon by an input with the URL ?

ericawright:
I see. I'd be willing to merge this without that enhancement, but yes, it would be more convenient to show the URL there. 
Design wise, I think it could work if it is in an container that is the length of the header, but no longer. And the text inside can scroll to select it all. then we don't stretch the table wider nor higher.
Fortunately very few people should actually have copy disabled. I'd be willing to merge this even if you don't want to go further and add that.

Johann-S:
I think if it's ok for you, you can merge this PR, and I'll make a new PR for this issue when I'll find time to work on that 👍

PR #412, opened 2017-08-03
throw key errors before download begins

### LINKED ISSUE #151: Confusing behavior with invalid secret key
Found in https://fileshare.stage.mozaws.net/__version__ (a7fcb1a):

```json
{
  "commit": "a7fcb1a",
  "source": "https://github.com/mozilla/something-awesome/",
  "version": "0.1.2"
}
```

---

### Steps to reproduce:
1. Go to https://fileshare.stage.mozaws.net/
1. Upload a file.
1. Copy the download URL.
1. Paste the URL into a new tab and shave a few characters off the `delete_token` in the URL. In my case, the random URL is https://fileshare.stage.mozaws.net/download/f9037eccaa5004e4c20db610044d0f95/#jcem2Ds6WLyL4HZ0yyagiA.
1. After you've submitted the [now broken] URL, click the _Download File_ button to download the file.

### Actual results:
The Dev Tools console says:
> The file has expired, or has already been deleted.

And
> Uncaught (in promise) TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
>    at fileReceiver.download.catch.then (bundle.js:1)
>    at <anonymous>

download_your_file

But looking at my downloaded files, it doesn't look like the file actually ever downloaded.

Going to the URL again, it seems like the download has been deleted from the server, but never actually successfully downloaded to my computer:

https://fileshare.stage.mozaws.net/download/f9037eccaa5004e4c20db610044d0f95/#jcem now says:

download_your_file

### Expected results:
- The file shouldn't be deleted off the server if it wasn't successfully downloaded.
- We should perhaps do better data validation and messaging to say things like "invalid delete_token" (although that probably doesn't mean anything to an end-user), or something that lets the user know that they maybe didn't copy the entire URL or something else happened.

dannycoates:
fixes #151 

This could use better messaging but at least with this the server won't delete the file before you can try the correct key. We should follow up with a messaging issue.

### DIFF
--- a/frontend/src/fileReceiver.js
+++ b/frontend/src/fileReceiver.js
@@ -7,42 +7,8 @@ class FileReceiver extends EventEmitter {
   }
 
   download() {
-    return Promise.all([
-      new Promise((resolve, reject) => {
-        const xhr = new XMLHttpRequest();
-
-        xhr.onprogress = event => {
-          if (event.lengthComputable && event.target.status !== 404) {
-            this.emit('progress', [event.loaded, event.total]);
-          }
-        };
-
-        xhr.onload = function(event) {
-          if (xhr.status === 404) {
-            reject(new Error('notfound'));
-            return;
-          }
-
-          const blob = new Blob([this.response]);
-          const fileReader = new FileReader();
-          fileReader.onload = function() {
-            const meta = JSON.parse(xhr.getResponseHeader('X-File-Metadata'));
-            resolve({
-              data: this.result,
-              aad: meta.aad,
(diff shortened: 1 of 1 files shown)


abhinadduri:
This is nice. How hard is it to localize any strings we would add? @flodolo @dannycoates

flodolo:
> This is nice. How hard is it to localize any strings we would add to the ftl files? @flodolo @dannycoates

I need context :-)

dannycoates:
follow up issue #413

PR #414, opened 2017-08-03
Add link to FAQ in unsupported view

### LINKED ISSUE #377: Improve "Your browser is not supported" message
Using current Safari

```
Your browser is not supported.
Unfortunately this browser does not support the web technology that powers Firefox Send. You’ll need to try another browser. We recommend Firefox!
```

I assume we're doing some kind of feature detection here -- and that this isn't a bug? -- and without wanting to create a ton of work, would it be possible to show a message that is more descriptive?

c.f., https://twitter.com/jiaaro/status/892474085262991360

pdehaan:
Just adds a link to the relevant FAQ bits, although we still incur an l10n hit since I had to add a string.
Totally looks like:

firefox_send

Fixes #377

### DIFF
--- a/views/unsupported.handlebars
+++ b/views/unsupported.handlebars
@@ -8,6 +8,7 @@
     </a>
   {{else}}
     <div class="description" data-l10n-id="notSupportedDetail">Unfortunately this browser does not support the web technology that powers Firefox Send. You’ll need to try another browser. We recommend Firefox!</div>
+    <div class="description"><a href="https://github.com/mozilla/send/blob/master/docs/faq.md#why-is-my-browser-not-supported" data-l10n-id="notSupportedLink" target="_blank" rel="noopener noreferrer"></a></div>
     <a id="dl-firefox" href="https://www.mozilla.org/firefox/new/?scene=2" target="_blank">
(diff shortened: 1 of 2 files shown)


views/unsupported.handlebars
```diff
@@ -8,6 +8,7 @@
     </a>
   {{else}}
     <div class="description" data-l10n-id="notSupportedDetail">Unfortunately this browser does not support the web technology that powers Firefox Send. You’ll need to try another browser. We recommend Firefox!</div>
+    <div class="description"><a href="https://github.com/mozilla/send/blob/master/docs/faq.md#why-is-my-browser-not-supported" data-l10n-id="notSupportedLink" target="_blank" rel="noopener noreferrer"></a></div>
```
ericawright: Can we put the English string inside the a-tags for IE and noscript scenarios

pdehaan: Absolutely...

Warning, the combination of `<noscript>` on the /unsupported/* route looks terrible (~~mostly~~ _entirely_ due to the fact my noscript hack doesn't seemingly conform to any style guide in the known universe).

firefox_send

PR #415, opened 2017-08-03
Only include Fira CSS on /unsupported/* route

### LINKED ISSUE #391: Fira is only used for unsupported-browser message
Cutting the custom font download could cut down load time and remove the extraneous `fira.css`.

pdehaan:
Fixes #391

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -96,7 +96,8 @@ app.get('/', (req, res) => {
 app.get('/unsupported/:reason', (req, res) => {
   const outdated = req.params.reason === 'outdated';
   res.render('unsupported', {
-    outdated: outdated
+    outdated,
+    fira: true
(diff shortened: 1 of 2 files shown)


ericawright:
Looks good. I also wonder if using Fira was intentional, since this is the only place it's used. But that's a question for @Pekochen, and in the meantime I see no problem with merging this.

pdehaan:
> Looks good. I also wonder if using Fira was intentional, since this is the only place it's used. But that's a question for @Pekochen, and in the meantime I see no problem with merging this.

I'd be more than happy to remove Fira. It feels a bit aggressive to display it only on that page. "Sorry your browser is unsupported, but here is an additional 170.75KB of fonts"...

firefox_send

Issue #411, opened 2017-08-03
Use webcrypto-liner polyfill to support Safari


jrmuizel:
It seems like it should be possible to support Safari using https://github.com/PeculiarVentures/webcrypto-liner

dannycoates:
This looks very cool. We'll look at it. Thanks for the suggestion!

dannycoates:
done!

Issue #398, opened 2017-08-02
Add target="_blank" to the "Try Send" link on the download page


johngruen:
Reported on twitter. If you click this link mid-download it messes the whole thing up!

pdehaan:
Mr. @gene1wood also had a related observation in https://github.com/mozilla/send/issues/401#issuecomment-320028143.
Not sure if the implication is that all footer links should have `target="_blank"` as well to avoid redirect jank.

PR #416, opened 2017-08-03
WIP: use webcrypto-liner to support Safari 10

dannycoates:


frontend/src/download.js
```diff
@@ -12,176 +12,186 @@ const Raven = window.Raven;
 
 $(document).ready(function() {
   //link back to homepage
```
pdehaan: This comment feels wrong now

dannycoates: hey man, WIP 😝

pdehaan: Free QA in Progress.

pdehaan:
Ref: #411

dannycoates:
There's one weird side effect on desktop Safari. After a successful download all the `target="_blank"` links *download the html* instead of opening a new tab. 🙈

views/download.handlebars
```diff
@@ -35,5 +35,5 @@
     </div>
   </div>
 
-  <a class="send-new" data-l10n-id="sendYourFilesLink" target="_blank"></a>
+  <a class="send-new" data-l10n-id="sendYourFilesLink" href="/"></a>
```
pdehaan: Didn't we have the `target="_blank"` there to prevent disrupting a download? #398

pdehaan:
What is "public/cryptofill.js"? It looks vendored, but not sure if could be copied from **asmcrypto.js** or **webcrypto-liner** devDeps at build time (like l20n and babel-polyfill).

views/download.handlebars
dannycoates: See https://github.com/mozilla/send/blob/master/frontend/src/links.js#L7-L19 which gets called at the start of a download.

dannycoates:
> What is "public/cryptofill.js"?

Its a custom build of those. I'll make that into a build step once we have something more sane to do it.

PR #368, opened 2017-08-01
WIP: Add babel+webpack scaffolding

pdehaan:
@dannycoates @ericawright I'm not a webpack expert, but this was as close as I got.
I think my bundles are working, but they implode if I try and do a production build using the `webpack -p` command, due to a `class` keyword in TestPilotGA.

```
/giphy "I have no idea what I'm doing"
```

PR #422, opened 2017-08-04
Move some devDependencies to normal dependencies

weihanglo:
This is a minor issue that would not affect development. However, if a 3rd library is used in actual source code, it is a good pattern to put it in **dependencies** IMHO.


weihanglo:
Close. See https://github.com/mozilla/send/pull/368#discussion_r131306089

PR #428, opened 2017-08-04
add twitter and open graph cards

### LINKED ISSUE #153: Add OpenGraph and Twitter Card metadata to pages
Minor nit, but will make the pages play much nicer with out other tools such as Activity Stream and Screenshots if users ever visit/screenshot a page on our site.

Refs:
- http://ogp.me/
- https://developers.facebook.com/tools/debug/
- https://dev.twitter.com/cards/types/summary
- https://cards-dev.twitter.com/validator

johngruen:
Fixes #153 

Note. I didn't localize b/c i'm not sure it's possible to localize content attrs.


views/layouts/main.handlebars
```diff
@@ -6,6 +6,16 @@
...
 
+  <meta property="og:title" content="Firefox Send"/>
+  <meta name="twitter:title" content="Firefox Send"/>
+  <meta name="description" content="Encrypt and send files with a link that automatically expires to ensure your important documents don’t stay online forever."/>
+  <meta property="og:description" content="Encrypt and send files with a link that automatically expires to ensure your important documents don’t stay online forever."/>
+  <meta name="twitter:description" content="Encrypt and send files with a link that automatically expires to ensure your important documents don’t stay online forever."/>
+  <meta name="twitter:card" content="summary"/>
+  <meta property="og:image" content="/resources/send-fb.jpg"/>
+  <meta name="twitter:image" content="/resources/send-twitter.jpg"/>
+  <meta property="og:url" content="https://send.firefox.com"/>
```
pdehaan: These last 3 seem almost questionable...

1. Should we be pulling `og:url` from some environment? Like check `conf.env` and set this to dev/stage/prod as appropriate? I'm not 100% certain what the `og:url` is used for.
2. For `og:image` and `twitter:image`, I thought those needed to be fully qualified URLs. I vaguely recall us struggling with this on Test Pilot site. Ref: [**mozilla/testpilot** /frontend/tasks/pages.js:16-19](https://github.com/mozilla/testpilot/blob/23a62b5a91042c46d20a317193332e4b167e1825/frontend/tasks/pages.js#L16-L19)

Although hard coding prod URLs was causing issues for QA since we cannot test the Twitter and Facebook cards until you push it to production, which isn't always ideal since it's too late at that point and we'd need more production pushes in order to verify tweaks.

Duplication of title/description strings doesn't seem DRY either. Not sure if we want to pass those into Handlebars as helpers (like we did for `availableLanguages`), not sure if that'd help w/ localization support or not since this would be server-side.

pdehaan: @dannycoates Does the Express server know whether it's serving as "https://send.firefox.com", "send.stage.mozaws.net", or "send.dev.mozaws.net"?

I think our only supported `NODE_ENV` values are "production", "development", and "test" (per [/server/config.js:35-39](https://github.com/mozilla/send/blob/80db158ee3de89cfe2de302814d7825d0a27fbd9/server/config.js#L35-L39)), so not sure I could use that to detect a difference between dev/stage.

pdehaan:
The weird thing is that these strings are _kinda_ already localized in the [/public/locales/*/send.ftl](https://github.com/mozilla/send/blob/455c4f5472b64efc6741cd0f3ebcf0c40b2d28ba/public/locales/en-US/send.ftl#L2-L6) files (but then we'd need to figure out server side translations):

```yaml
title = Firefox Send
uploadPageExplainer = Send files through a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
```

Also our meta tags and site copy vary by _just_ a few words... Super ultra nit, but possibly worth pointing out. FWIW, I like your new copy better (mainly because I don't care for the word "stuff"), but changing the FTL files to match new copy would have a l10n penalty:

### Current `uploadPageExplainer` FTL text:

> "Send files through a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever."

### Proposed `<meta description="">` text:

> "Encrypt and send files with a link that automatically expires to ensure your important documents don’t stay online forever."

johngruen:
@pdehaan i used the copy from the test pilot site which, for reasons i don't fully understand, makes legal more happy.

server/server.js
```diff
@@ -38,6 +38,16 @@ function prodLangs() {
 
 const availableLanguages = conf.l10n_dev ? allLangs() : prodLangs();
 
+const envURL = (env) => {
```
johngruen: using this to set the URL so this can be tested on dev environments. Not super familiar with convict, but maybe this would be better done in config

server/server.js
```diff
@@ -58,7 +68,10 @@ app.engine(
     partialsDir: 'views/partials/',
     helpers: {
       availableLanguages,
-      l10nDev: conf.l10n_dev
+      l10nDev: conf.l10n_dev,
+      envURL: envURL(conf.env),
+      title: 'Firefox Send',
+      description: 'Encrypt and send files with a link that automatically expires to ensure your important documents don’t stay online forever.'
```
johngruen: Note, this doesn't get localized, maybe that's not possible...

server/server.js
```diff
@@ -38,6 +38,16 @@ function prodLangs() {
 
 const availableLanguages = conf.l10n_dev ? allLangs() : prodLangs();
 
+const envURL = (env) => {
+  switch (env) {
+    case 'test':
+      return 'https://send.stage.mozaws.net';
```
pdehaan: I'm not convinced this is a thing.
I have a feeling that stage would use an env of "production". Not sure what [if anything] uses the "test" NODE_ENV, unless Circle-CI somehow sets it to that, but I strangely kind of doubt it.

I don't know how to tell the detect whether we're running on stage vs prod on the server.

dannycoates:
I extracted the url to a env var. Here's the issue to set it https://github.com/mozilla-services/cloudops-deployment/issues/1018

PR #433, opened 2017-08-04
Prevent download button from being clicked multiple times

### LINKED ISSUE #429: If the "Download" button is clicked multiple times, the file is wrongly downloaded for each click
[Notes]:
- The pregressbar and the percentage are also displayed for each download and are overlapped.

[Affected versions]:
- Firefox 54.0.1 and up

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Steps to reproduce]:
1. Start the Firefox browser and navigate to "https://send.stage.mozaws.net/" page.
2. Upload a file and copy the link.
3. Navigate to the file download page.
4. Multiple click on the "Download" button.

[Expected result]:
- You are redirected to the "Downloading" page and the file is downloaded once.

[Actual results]:
- You are redirected to the "Downloading" page, but the file is downloaded for each click.

[Additional Notes]:
- Probably this issue can be easily encountered when you have a bad connection.  
- The issue is only reproducible on dev and stage servers.
- Here is a screen recording with the issue:
multiple downloads

pdehaan:
This seems to work locally for me, but it happens so quickly sometimes I have a hard time clicking fast enough.

Fixes #429

### DIFF
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -42,6 +42,9 @@ $(document).ready(function() {
   });
   $('#download-btn').click(download);
   function download() {
+    // Disable the download button to avoid accidental double clicks.
+    $('#download-btn').attr('disabled', 'disabled');
+
     storage.totalDownloads += 1;
 
     const fileReceiver = new FileReceiver();


frontend/src/download.js
```diff
@@ -42,6 +42,9 @@ $(document).ready(function() {
   });
   $('#download-btn').click(download);
   function download() {
+    // Disable the download button to avoid accidental double clicks.
+    $('#download-btn').attr('disabled', 'disabled');
```
Johann-S: Maybe we can store this selector (`#download-btn`) in a var because it seems we try to retrieve this ID a few times
so something like this would be great : 
```javascript
const $downloadBtn = $('#download-btn');
```

Issue #284, opened 2017-07-24
Cache jQuery lookups and chain some commands?


pdehaan:
While doing some PRs, I noticed we have a few places that may be micro-optimized by caching lookups and chaining some stuff.

Random snippet:
```js
      window.setTimeout(() => {
        const linkImg = document.createElement('img');
        $(linkImg).addClass('icon-copy');
        $(linkImg).attr('data-l10n-id', 'copyUrlHover');
        $(linkImg).attr('src', '/resources/copy-16.svg');
        $(link).html(linkImg);
      }, 500);
```

We could probably do:

```js
$(linkImg).addClass('icon-copy')
  .attr({
    'data-l10n-id': 'copyUrlHover',
    'src': '/resources/copy-16.svg'
  });
```

Similarly, caching some jQuery lookups may shave a few ms:

```js
  $('.upload-window').on('dragover', () => {
    $('.upload-window').addClass('ondrag');
  });
  $('.upload-window').on('dragleave', () => {
    $('.upload-window').removeClass('ondrag');
  });
```

Maybe somewhere in `$(document).ready(() => {`, we can cache `const $uploadWindow = $('.upload-window');`, then do:

```js
$uploadWindow.on('dragover', () => {
  $uploadWindow.addClass('ondrag');
}).on('dragleave', () => {
  $uploadWindow.removeClass('ondrag');
});
```

Johann-S:
I did the last change on my PR here : https://github.com/mozilla/send/pull/409/files#diff-cb458696a7cf1f8badab576ac8ccc6afR113

PR #444, opened 2017-08-04
Chain jQuery calls, do not use events alias and store selectors

Johann-S:
This PR add some improvements about perfs and good practices.

- Remove events shortcuts ( `.click` transformed to `.on('click')`
- Chain jQuery calls
- Store selectors to avoid DOM access when we had already find the DOM elements
- Change `$(document).ready` to `$(function () { })` with jQuery 3 it's recommended to use this form to listen the DOM ready events (see : https://api.jquery.com/ready/)

Fix : https://github.com/mozilla/send/issues/284

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -26,21 +26,30 @@ const allowedCopy = () => {
   return support ? document.queryCommandSupported('copy') : false;
 };
 
-$(document).ready(function() {
+$(() => {
   gcmCompliant()
-    .then(function() {
-      $('#page-one').removeAttr('hidden');
-      $('#file-upload').change(onUpload);
+    .then(function () {
+      const $pageOne = $('#page-one');
+      const $copyBtn = $('#copy-btn');
+      const $link = $('#link');
+      const $uploadWindow = $('.upload-window');
+      const $ulProgress = $('#ul-progress');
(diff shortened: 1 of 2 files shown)


frontend/src/upload.js
```diff
@@ -81,28 +103,27 @@ $(document).ready(function() {
...
       $copyBtn.attr('disabled', true);
-      $('#link').attr('disabled', true);
+      $link.attr('disabled', true);
       $copyBtn.html(
         '<img src="/resources/check-16.svg" class="icon-check"></img>'
       );
       window.setTimeout(() => {
         $copyBtn.attr('disabled', false);
-        $('#link').attr('disabled', false);
+        $link.attr('disabled', false);
```
pdehaan: You could potentially move `$link.attr()` up a line and chain both `$copyBtn.attr()` calls.
I think you can also set multiple values in a single `.attr()` by passing an object: http://api.jquery.com/attr/#attr-attributes

```js
$link.attr('disabled', false);
$copyBtn.attr({
  disabled: false,
  'data-l10n-id': 'copyUrlFormButton'
});
```

Johann-S: changed that thank you @pdehaan 👍

dannycoates:
Thanks @Johann-S 👍

PR #456, opened 2017-08-06
Cater for mobile and desktop

### LINKED ISSUE #421: "Select a file on your computer" looks weird on mobile
Perhaps "Select a file on this device"?

Ref: [/public/locales/en-US/send.ftl:10-11](https://github.com/mozilla/send/blob/4ac07800c409f3c73e22b73b97ff5255d029e068/public/locales/en-US/send.ftl#L10-L11)

hubdotcom:
Fixes #421


public/locales/en-US/send.ftl
```diff
@@ -7,8 +7,8 @@ uploadPageExplainer = Send files through a safe, private, and encrypted link tha
 uploadPageLearnMore = Learn more
 uploadPageDropMessage = Drop your file here to start uploading
 uploadPageSizeMessage = For the most reliable operation, it’s best to keep your file under 1GB
-uploadPageBrowseButton = Select a file on your computer
-    .title = Select a file on your computer
+uploadPageBrowseButton = Select a file to upload
+    .title = Select a file to upload
```
flodolo: This is the process to change existing strings.
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

You would need a new string ID.

hubdotcom:
A string ID change is probably beneficial to ensure that translators update their definitions too..

PR #457, opened 2017-08-06
factored out progress into progress.js

### LINKED ISSUE #349: l10n for .progress-text
The variables order and "of" text.

### LINKED ISSUE #366: Make sure file sizes are localized
Subtle detail from https://github.com/mozilla/send/issues/349#issuecomment-318695786:

> It should be fully localized to allow reordering, something like
> 
> ```js
> // This message is displayed when uploading a file, e.g. "1,3 MB of 10 MB".
> uploadStatusMessage = { $size_loaded } of { $size_total }
> ```

Note how the comment says "1,3 MB". Currently we're using [**bytes**](http://npm.im/bytes) to calculate file sizes for [upload](https://github.com/mozilla/send/blob/dc83ec112362adac703527d9e326205fb5b7baeb/frontend/src/upload.js#L195-L200) and [download](https://github.com/mozilla/send/blob/dc83ec112362adac703527d9e326205fb5b7baeb/frontend/src/download.js#L75-L79), but I'm not sure that it supports localizations.

### LINKED ISSUE #462: File downloads break if you have a double quote in the filename
Randomly spotted in https://github.com/mozilla/send/pull/457/files#r131576048 but reproduced in production...

### Steps to reproduce:

1. Go to https://send.firefox.com
1. Upload a file with a double quote in the filename (for example, `batmanapp"roves.gif` below):

    batmanapp roves

1. Try and download that file on Send.

### Actual results:

firefox_send

```html
<div class="title">
      <span id="dl-filename" data-l10n-id="downloadFileName" data-l10n-args="{&quot;filename&quot;: &quot;batmanapp&quot;roves.gif&quot;}"></span>
      <span data-l10n-id="downloadFileSize" data-l10n-args="{&quot;size&quot;: &quot;1.01MB&quot;}"></span>
      <span id="dl-bytelength" hidden="true">1064054</span>
      <span id="dl-ttl" hidden="true">85626000</span>
    </div>
```

dannycoates:
Fixes #349 
Fixes #366
Fixes #462

### DIFF
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -1,144 +1,118 @@
 const { Raven } = require('./common');
 const FileReceiver = require('./fileReceiver');
-const { notify, gcmCompliant } = require('./utils');
-const bytes = require('bytes');
+const { bytes, notify, gcmCompliant } = require('./utils');
 const Storage = require('./storage');
 const storage = new Storage(localStorage);
 const links = require('./links');
 const metrics = require('./metrics');
-
+const progress = require('./progress');
 const $ = require('jquery');
-require('jquery-circle-progress');
+
+function onUnload(size) {
+  metrics.cancelledDownload({ size });
+}
+
+function download() {
+  const $downloadBtn = $('#download-btn');
+  const $title = $('.title');
+  const $file = $('#dl-file');
+  const size = Number($file.attr('data-size'));
+  const ttl = Number($file.attr('data-ttl'));
+  const unloadHandler = onUnload.bind(null, size);
+  const startTime = Date.now();
+  const fileReceiver = new FileReceiver();
+
+  $downloadBtn.attr('disabled', 'disabled');
+  $('#download-page-one').attr('hidden', true);
+  $('#download-progress').removeAttr('hidden');
+  metrics.startedDownload({ size, ttl });
+  links.setOpenInNewTab(true);
+  window.addEventListener('unload', unloadHandler);
+
+  fileReceiver.on('progress', data => {
+    progress.setProgress({ complete: data[0], total: data[1] });
+  });
+
+  let downloadEnd;
+  fileReceiver.on('decrypting', () => {
+    downloadEnd = Date.now();
+    window.removeEventListener('unload', unloadHandler);
+    fileReceiver.removeAllListeners('progress');
+    document.l10n.formatValue('decryptingFile').then(progress.setText);
+  });
+
+  fileReceiver.on('hashing', () => {
+    document.l10n.formatValue('verifyingFile').then(progress.setText);
+  });
+
+  fileReceiver
+    .download()
+    .catch(err => {
+      metrics.stoppedDownload({ size, err });
+
+      if (err.message === 'notfound') {
+        location.reload();
+      } else {
+        document.l10n.formatValue('errorPageHeader').then(translated => {
+          $title.text(translated);
+        });
+        $downloadBtn.attr('hidden', true);
+        $('#expired-img').removeAttr('hidden');
+      }
+      throw err;
+    })
+    .then(([decrypted, fname]) => {
+      const endTime = Date.now();
+      const time = endTime - startTime;
+      const downloadTime = endTime - downloadEnd;
+      const speed = size / (downloadTime / 1000);
+      storage.totalDownloads += 1;
+      metrics.completedDownload({ size, time, speed });
+      progress.setText(' ');
+      document.l10n
+        .formatValues('downloadNotification', 'downloadFinish')
+        .then(translated => {
+          notify(translated[0]);
+          $title.text(translated[1]);
+        });
+
+      const dataView = new DataView(decrypted);
+      const blob = new Blob([dataView]);
+      const downloadUrl = URL.createObjectURL(blob);
+
+      const a = document.createElement('a');
+      a.href = downloadUrl;
+      if (window.navigator.msSaveBlob) {
+        // if we are in microsoft edge or IE
+        window.navigator.msSaveBlob(blob, fname);
+        return;
+      }
+      a.download = fname;
+      document.body.appendChild(a);
+      a.click();
+    })
+    .catch(err => {
+      Raven.captureException(err);
+      return Promise.reject(err);
+    })
+    .then(() => links.setOpenInNewTab(false));
+}
 
 $(() => {
+  const $file = $('#dl-file');
+  const filename = $file.attr('data-filename');
+  const b = Number($file.attr('data-size'));
+  const size = bytes(b);
+  document.l10n
+    .formatValue('downloadFileSize', { size })
+    .then(str => $('#dl-filesize').text(str));
+  document.l10n
+    .formatValue('downloadingPageProgress', { filename, size })
+    .then(str => $('#dl-title').text(str));
+
   gcmCompliant()
     .then(() => {
-      const $downloadBtn = $('#download-btn');
-      const $dlProgress = $('#dl-progress');
-      const $progressText = $('.progress-text');
-      const $title = $('.title');
-
-      const filename = $('#dl-filename').text();
-      const size = Number($('#dl-size').text());
-      const ttl = Number($('#dl-ttl').text());
-
-      //initiate progress bar
-      $dlProgress.circleProgress({
-        value: 0.0,
-        startAngle: -Math.PI / 2,
-        fill: '#3B9DFF',
-        size: 158,
-        animation: { duration: 300 }
-      });
-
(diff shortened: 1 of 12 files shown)


frontend/src/fileReceiver.js
```diff
@@ -62,7 +62,7 @@ class FileReceiver extends EventEmitter {
         });
       })
       .then(([fdata, key]) => {
-        this.emit('decrypting', true);
+        this.emit('decrypting');
```
dannycoates: Turns out we don't really need the start/finish part of these events. Removing the finished case simplifies the listeners.

public/locales/en-US/send.ftl
```diff
@@ -50,6 +51,7 @@ downloadButtonLabel = Download
     .title = Download
 downloadNotification = Your download has completed.
 downloadFinish = Download Complete
+fileSizeProgress = ({ $partialSize } of { $totalSize })
```
dannycoates: This PR adds a couple strings for l10n, but doesn't use them yet. @flodolo, is it a good idea to pre-stage strings so they can be translated before using them, or better just to use them now? 

In `fileSizeProgress` case it's a new string previously not localized, whereas `uploadingPageProgress` will replace the `uploadingPageHeader` to be symmetric with `downloadPageProgress`.

frontend/src/download.js
```diff
@@ -1,144 +1,108 @@
...
+      return Promise.reject(err);
+    })
+    .then(() => links.setOpenInNewTab(false));
+}
 
-      $downloadBtn.on('click', download);
+$(() => {
+  gcmCompliant()
+    .then(() => {
+      $('#download-btn').on('click', download);
```
dannycoates: bumped `download()` up to make the "main" function cleaner.

public/locales/en-US/send.ftl
flodolo: In general it's not a good idea to preland strings, because a) you won't have any way to test it and b) you might have to change your mind in the implementation phase.

Note: if this PR makes a string obsolete, that string should be removed.

dannycoates:
now also fixes #349

dannycoates:
and fixes #366

pdehaan:
Off-topic, but I think the "Fixes #___" only auto-closes the referenced bugs if you include it in the commit message or original PR comment.

frontend/src/utils.js
```diff
@@ -104,9 +104,29 @@ function copyToClipboard(str) {
   return result;
 }
 
+const LOCALIZE_NUMBERS = !!(
+  typeof Intl === 'object' &&
+  Intl &&
+  typeof Intl.NumberFormat === 'function'
+);
+
+const UNITS = ['B', 'kB', 'MB', 'GB'];
```
pdehaan: Does this work for l10n? Is a "MB" the same in English, German, Japanese, Russian, etc?
/cc @flodolo @mathjazz

frontend/src/utils.js
```diff
@@ -104,9 +104,29 @@ function copyToClipboard(str) {
...
+function bytes(num) {
+  const exponent = Math.min(Math.floor(Math.log10(num) / 3), UNITS.length - 1);
+  const n = Number(num / Math.pow(1000, exponent));
+  const nStr = LOCALIZE_NUMBERS
+    ? n.toLocaleString(navigator.language, {
+        minimumFractionDigits: 1,
+        maximumFractionDigits: 1
+      })
```
pdehaan: Sorry, I've always struggled to understand localization...
Not sure I understand how this is working... What happens here if I have a local `navigator.language` of some non-supported localization on stage/prod (like Hebrew, for example), but I get a match on my second or third choice l20n fallback (such as Japanese)? Would the numbers be displayed in my preferred Hebrew language format, but the surrounding text be Japanese because that is one of the official production `availableLanguages`? Or am I totally missing something?

views/download.handlebars
```diff
@@ -2,13 +2,13 @@
   <script src="/download.js"></script>
   <div id="download-page-one">
     <div class="title">
-      <span id="dl-filename"
+      <span id="dl-file"
+            data-filename="{{filename}}"
+            data-size="{{sizeInBytes}}"
+            data-ttl="{{ttl}}"
             data-l10n-id="downloadFileName"
             data-l10n-args='{"filename": "{{filename}}"}'></span>
```
pdehaan: Unrelated to this glorious PR, but looking at it with my special eyes, I think this is a bug...
I can seemingly break the download page hard on production by uploading a file with a pair of double quotes in the filename. Not sure if we can encode the filename here (and on line 6 above, and possibly elsewhere, or what.

frontend/src/utils.js
flodolo: @zbraniecki or @stas
Any advice?

frontend/src/utils.js
flodolo: We localize them, and you don't even need to go to a different script
https://transvision.mozfr.org/string/?entity=toolkit/chrome/mozapps/downloads/downloads.properties:megabyte&repo=central

Having said that, I wonder if we have access to them through Intl API @zbraniecki would know for sure.

Side note: kB is not going to be consistent with Firefox en-US (don't get me started)
https://hg.mozilla.org/mozilla-central/file/default/toolkit/locales/en-US/chrome/mozapps/downloads/downloads.properties#l67

zbraniecki: Unfortunately, the https://github.com/zbraniecki/proposal-intl-unit-format has not been implemented yet in ECMA402 :(

frontend/src/utils.js
zbraniecki: First of all, never use `navigator.language`, use `navigator.languages` - this is a fallback chain of your preferred languages.

Now, what do you mean by "my second or third choice l20n fallback"? If I understand correctly this is a website, so it will get `navigator.languages` as a fallback chain from the user.
You just have to pass it to `toLocaleString` and it'll pick the best locale matching the list.

views/download.handlebars
dannycoates: Nice catch @pdehaan. We ought to be doing `data-l10n-args='{{JSON.stringify({filename})}}'`

frontend/src/utils.js
pdehaan: @zbraniecki Sorry, I was trying to explain that I only speak Hebrew, Japanese, and Bulgarian:

preferences

I wasn't sure if this meant that my file sizes would be in Hebrew, but the rest of the site would be Japanese (or Bulgarian, if Japanese wasn't an available language for Firefox Send site + l20n).

zbraniecki: > I wasn't sure if this meant that my file sizes would be in Hebrew (which is not an available language for Send), but the rest of the site would be Japanese (or Bulgarian, if Japanese wasn't an available language for Firefox Send site + l20n).

Ah, yes. If you pass `navigator.languages` it will be `["he", "ja", 'bg"]`. Since all browsers carry NumberFormat data for `he`, we'll format into `he`.

If you want to link the raw Intl API to use the same languages as were negotiated for l20n, you need to retrieve the result of language negotiation for l20n.

I'm not sure how this API currently works, but it should be something along the lines of `document.l10n.locales` which should be the list of locales that l20n negotiated down to.
Since `he` is not available, the list would end up being sth like `['ja', 'bg', 'en-US']` assuming both `ja` and `bg` are available, and `en-US` is the last fallback.

In that case, passing that list fo the Intl API will do the right thing for you.

I haven't been working with `l20n.js` package in a while, so CC'ing @stasm, who's maintaining it - he will hopefully be able to point out how to retrieve the negotiated locale list from l20n to pass to the Intl API.

views/download.handlebars
pdehaan: Possibly a few places:

```sh
$ git grep -n "{{filename}}"

views/download.handlebars:7:       data-l10n-args='{"filename": "{{filename}}"}'></span>
views/download.handlebars:23:      data-l10n-args='{"filename": "{{filename}}", "size": "{{filesize}}"}'>
views/download.handlebars:34:      <div class="progress-text">{{filename}}</div>
```

frontend/src/fileReceiver.js
abhinadduri: Although I don't think it affects the outcome of the tests, I think we should change the frontend tests to reflect that these listeners only get called once.

frontend/src/utils.js
stasm: Sorry, I missed the notification for this. `l20n.js` exposes the currently negotiated languages in the `langs` attribute of the `<html>` element. You can access it with `document.documentElement.getAttribute('langs')`. It's a space-separated list of local codes.

PR #466, opened 2017-08-07
added webpack

dannycoates:
This is just a start at webpack and transpiling to es2015. 

@pdehaan I excluded the "env" preset for now for simplicity. We should revisit it along with removing the full polyfill.min.js script from the template in a follow up issue.

Bonus: fixes Edge (I think)

### DIFF
--- a/frontend/src/links.js
+++ b/frontend/src/links.js
@@ -1,7 +1,7 @@
 let links = [];
 
 document.addEventListener('DOMContentLoaded', function() {
-  links = document.querySelectorAll('a:not([target])');
+  links = Array.from(document.querySelectorAll('a:not([target])'));
 });
 
 function setOpenInNewTab(bool) {
--- a/frontend/src/metrics.js
+++ b/frontend/src/metrics.js
@@ -1,4 +1,4 @@
-const testPilotGA = require('testpilot-ga');
+const testPilotGA = require('testpilot-ga/src/TestPilotGA');
 const Storage = require('./storage');
 const storage = new Storage(localStorage);
 
@@ -196,7 +196,7 @@ function exitEvent(target) {
 }
 
 function addExitHandlers() {
-  const links = document.querySelectorAll('a');
+  const links = Array.from(document.querySelectorAll('a'));
   links.forEach(l => {
     if (/^http/.test(l.href)) {
       l.addEventListener('click', exitEvent);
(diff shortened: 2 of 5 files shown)


frontend/src/links.js
```diff
@@ -1,7 +1,7 @@
 let links = [];
 
 document.addEventListener('DOMContentLoaded', function() {
-  links = document.querySelectorAll('a:not([target])');
+  links = Array.from(document.querySelectorAll('a:not([target])'));
```
dannycoates: Edge doesn't have `forEach` on `NodeList`

frontend/src/metrics.js
```diff
@@ -1,4 +1,4 @@
-const testPilotGA = require('testpilot-ga');
+const testPilotGA = require('testpilot-ga/src/TestPilotGA');
```
dannycoates: This is a bit hacky but it lets us transpile the lib how we want to

package.json
```diff
@@ -17,7 +17,12 @@
   },
   "devDependencies": {
     "asmcrypto.js": "0.0.11",
+    "babel-core": "^6.25.0",
+    "babel-loader": "^7.1.1",
+    "babel-plugin-add-module-exports": "^0.2.1",
     "babel-polyfill": "^6.23.0",
+    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-stage-2": "^6.24.1",
     "browserify": "^14.4.0",
```
dannycoates: we should follow up to remove this, but i didn't want to mess with `test--browser` yet

frontend/src/metrics.js
pdehaan: It technically _works_, but only due to a bug in https://github.com/mozilla/testpilot-ga/pull/13.
Not sure how you want to address. We'd either have to re-compile the **testpilot-ga** dist files, or kill my upstream PR and remove the bad `files` in **testpilot-ga** repo.

dannycoates: we can cross that bridge when we get there ;)

pdehaan: Totally, just wanted to point out that due to the fact of loose dependency versions (`"testpilot-ga": "^0.3.0"`), semver, and upstream repos, that could technically break at any point in the future when @chuckharmston publishes a new version of **testpilot-ga**.

PR #469, opened 2017-08-07
Increase mimimum node version to 8.2.0

### LINKED ISSUE #467: May be vulnerable to header splitting
Send currently requires  "node": ">=8.0.0"

Version 8.2.0 https://github.com/nodejs/node/commit/e10c75579b536581ddd7ae4e2c3bf8a9d550d343 Added some protection against header splitting.

See http://jsgeek.com/posts/node-response-splitting.html and https://www.owasp.org/index.php/CRLF_Injection for further information on why this is required.

All data used with http headers (via writeHead etc) MUST be sanitised, at a minimum removing any non-printable characters (Carriage return, line feed, tab etc) either explicitly or perhaps by relying upon the 'new' behaviour in 8.2.0 to do this by default.

For example, https://github.com/mozilla/send/blob/c91d24cd86515f162b0f6731902a004ca5087fa6/server/server.js#L173 looks problematic.

I would suggest the a minimum version is increase to 8.2.0, as this may prevent the issue in with changes in the future if it is not considered by the author. Sanitising all data that will be used with writeHead (etc) would be a suitable alternative (by itself) if this cannot be done yet.

ehuggett:
This version has a useful protection against http header splitting

fixes #467

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "webpack": "^3.4.1"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=8.2.0"
   },
   "homepage": "https://github.com/mozilla/send/",
   "license": "MPL-2.0",


dannycoates:
Thank you @ehuggett 😄

Issue #440, opened 2017-08-04
An “img” element must have an “alt” attribute, except under certain conditions


pdehaan:
via #139 

```
line 47   col 26  Error  An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.  html-checker
line 109  col 3   Error  An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.  html-checker
line 116  col 57  Error  An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.  html-checker
line 124  col 91  Error  An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.  html-checker
line 125  col 91  Error  An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.  html-checker
```

Possibly related/duplicated:

```
line 112  col 57  Error  Images must have alternate text                        axe
line 120  col 91  Error  Images must have alternate text                        axe
line 121  col 91  Error  Images must have alternate text                        axe
```

This may be a bit trickier to debug, since we're setting some localized `alt=""` tags via l20n.

urohit011:
I would like to work on this.

pdehaan:
@urohit011 That would be awesome, thanks. 👍
Let us know if you have any questions.

urohit011:
Can you please point me to one of the files that has to be fixed.

varghesethomase:
You can find the images in main.handlebars. The images that does not have alt attribute are located at the bottom fo the page ie mozilla logo to the left and twitter and github logo on the right @urohit011

urohit011:
Thank you @varghesethomase

dannycoates:
fixed in #474

Issue #438, opened 2017-08-04
Text of buttons and links should not be repeated in the image alternative


pdehaan:
via #139 

```
line 21   col 7   Error  Text of buttons and links should not be repeated in the image alternative                                                                                 axe
```

Not entirely sure I know what duplicate `alt=""` this is talking about, but we're doing a bit of magic with l20n. Even w/ cURL and DOM inspectors, I can't seem to find a line 21 that matches anything remotely close to an `alt=""` or `title=""` attribute.

varghesethomase:
The bug here is the alt attribute for image has text "Send" and the text wrapped by the h1 tag is again "Send". <img src="/resources/send_logo.svg" alt="Send"/><h1 class="site-title">Send</h1>

varghesethomase:
If you can get me the verbiages to be used, I will take this up.

dannycoates:
fixed in #474

PR #474, opened 2017-08-08
Fixing bug #438 by adding role attribute to anchor tags and alt attribute images

varghesethomase:
Solves #438 and together solves #440.


dannycoates:
Thank you @varghesethomase 😄

PR #481, opened 2017-08-08
Cater for mobile and desktop

### LINKED ISSUE #421: "Select a file on your computer" looks weird on mobile

pdehaan:
@flodolo Does this look right?
Not sure how the other locales will handle the missing new ID, if they'll have an en_US fallback, or if the button label will say "uploadPageBrowseButton1" until the localizations come in.

Based on #456
Fixes #421


flodolo:
> Not sure how the other locales will handle the missing new ID, if they'll have an en_US fallback, or if the button label will say "uploadPageBrowseButton1" until the localizations come in.

They should fall back to the ID, as far as I know, which means we should wait for at least major locales to catch up. One possible solution is to land the new string separately, then wait to merge this one (using the string) later.

Issue #174, opened 2017-07-07
[enhancement] Look into RFC 8188


clouserw:
Martin Thomson commented on our flow diagram about his work in RFC 8188 to chunk large encrypted files over HTTP.  Let's look into that here after we ship v1.  Benefits would be:

  - Client could start decrypting sooner
  - Client wouldn't need to download the entire file to learn the first byte was bad
  - We could do some more complex resume operations if a download is interrupted
  - Combined with P2P there could be some interesting scenarios and speedups

ehuggett:
Would chunking (in theory) also make it possible for the receiver to start downloading before the upload has been completed by the sender? As files expire after 24 hours this might be useful for sending larger files when both parties have a slow connection.

dannycoates:
Yes, this could (in theory) allow the download and upload to happen in parallel. Another advantage is that it uses less memory so would allow for larger files and more limited devices.

PR #492, opened 2017-08-10
make the site mostly work when cookies (localStorage) are disabled

dannycoates:
In Firefox when cookies are disabled so is localStorage. The current version 1.1.0 is completely unusable when this is set and I've noticed a few related errors on sentry. 

This PR makes uploading and downloading work, but the file list (as expected) does not.

### DIFF
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -2,7 +2,7 @@ const { Raven } = require('./common');
 const FileReceiver = require('./fileReceiver');
 const { bytes, notify, gcmCompliant } = require('./utils');
 const Storage = require('./storage');
-const storage = new Storage(localStorage);
+const storage = new Storage();
 const links = require('./links');
 const metrics = require('./metrics');
 const progress = require('./progress');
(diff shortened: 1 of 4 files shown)


pdehaan:
Q: Should we display some notification about "enhanced" file list functionality when a user has cookies disabled?

dannycoates:
> Should we display some notification about "enhanced" file list functionality when a user has cookies disabled?

I think this would be a good follow-up issue for UX. There's other scenarios where localStorage isn't available as well. The one I'm thinking of is Firefox Focus. In that particular case maybe we want to assume the user doesn't expect to be able to see upload history at all, so notifying them would just be annoying?

dannycoates:
Speaking of Focus, this PR makes uploading work for me on Focus for iOS, but download is still broken.

ehuggett:
As the fallback storage eventually falls back to "map()" i presume this causes more issues as the file gets larger? (up to the point were it is unusable due to the memory consumption or the browser actually won't allow it)

From a quick search it appears to be possible to read a local file in memory in "chunks", which could be encrypted and uploaded before loading the next one. Files would presumably have to be uploaded slightly differently to accommodate this (or can you append data onto a HTTP post that is already in-progress? I assume not, so multiple POST's/etc)

The idea came from https://github.com/mozilla/send/issues/174

PR #493, opened 2017-08-10
added webpack-dev-middleware

dannycoates:
Now `npm start` or `npm run dev` will rebuild js bundles when the source changes on localhost.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -38,6 +38,17 @@ function prodLangs() {
 
 const availableLanguages = conf.l10n_dev ? allLangs() : prodLangs();
 
+if (conf.env === 'development') {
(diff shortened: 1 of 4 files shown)


server/server.js
```diff
@@ -38,6 +38,17 @@ function prodLangs() {
 
 const availableLanguages = conf.l10n_dev ? allLangs() : prodLangs();
 
+if (conf.env !== 'production') {
```
pdehaan: Just to confirm, the send.stage.mozaws.net server runs in a NODE_ENV="production" environment, correct?

dannycoates: yeah, but `=== 'development'` is better

PR #499, opened 2017-08-10
use import/export in the frontend code

dannycoates:
i guess this is the future?

### DIFF
--- a/frontend/src/download.js
+++ b/frontend/src/download.js
@@ -1,13 +1,13 @@
-const { Raven } = require('./common');
-const FileReceiver = require('./fileReceiver');
-const { bytes, notify, gcmCompliant } = require('./utils');
-const Storage = require('./storage');
-const storage = new Storage();
-const links = require('./links');
-const metrics = require('./metrics');
-const progress = require('./progress');
-const $ = require('jquery');
+import { Raven } from './common';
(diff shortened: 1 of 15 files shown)


frontend/src/download.js
```diff
@@ -1,13 +1,13 @@
...
+import { Raven } from './common';
+import FileReceiver from './fileReceiver';
+import { bytes, notify, gcmCompliant } from './utils';
+import Storage from './storage';
+import * as links from './links';
+import * as metrics from './metrics';
+import * as progress from './progress';
+import $ from 'jquery';
 
+const storage = new Storage(localStorage);
```
pdehaan: Does the `Storage()` constructor still take arguments? I thought you changed that in #492...

frontend/src/upload.js
```diff
@@ -1,19 +1,19 @@
...
-const metrics = require('./metrics');
-const progress = require('./progress');
+} from './utils';
+import Storage from './storage';
+import * as metrics from './metrics';
+import * as progress from './progress';
+import $ from 'jquery';
 
-const $ = require('jquery');
+const storage = new Storage(localStorage);
```
pdehaan: Same as above. Should this be `new Storage(localStorage)` or just `new Storage()`?

PR #502, opened 2017-08-11
extracted filelist into its own file

dannycoates:


frontend/src/upload.js
```diff
@@ -11,15 +12,11 @@ import {
 import Storage from './storage';
 import * as metrics from './metrics';
 import * as progress from './progress';
+import * as fileList from './fileList';
```
pdehaan: ESLint sayeth:

```
/home/ubuntu/send/frontend/src/upload.js
  15:13  error    'fileList' is defined but never used  no-unused-vars
```

frontend/src/utils.js
```diff
@@ -129,9 +129,15 @@ function percent(ratio) {
     : `${Math.floor(ratio * 100)}%`;
 }
 
+function allowedCopy() {
+  const support = !!document.queryCommandSupported;
+  return support ? document.queryCommandSupported('copy') : false;
+};
```
pdehaan:
```
/home/ubuntu/send/frontend/src/utils.js
  135:2  error  Unnecessary semicolon  no-extra-semi
```

frontend/src/storage.js
```diff
@@ -86,7 +86,10 @@ export default class Storage {
   }
 
   getFileById(id) {
-    return this.engine.getItem(id);
+    try {
+    return JSON.parse(this.engine.getItem(id));
```
pdehaan: Indenting here looks v wrong. If I had to guess, it's because we've disabled indenting checks in ESLint since that's handled automatically by **prettier**. 

One solution may be to have some auto-format+lint code on a Git hook. It'd force **prettier** formatting on each commit, which may cause 1-2 noisy commits w/ unrelated formatting fixes at first, but then probably 👌 after that initial hiccup.

Issue #306, opened 2017-07-25
[Enhancement] The cursor should have a Ø symbol in order to better represent the invalid action 


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 54.0 and up

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Steps to reproduce]:
1. Open the browser and navigate to https://send.dev.mozaws.net/.
2. Select a text and drag it on the Drop area.
3. Observe the cursor.

[Expected result]:
- The cursor is changed and shows a Ø symbol.

[Actual results]:
- Nothing happens.

[Notes]:
- As it stands now, some users might not understand why dropping text (or links, images) does not work, because there is no feedback for them.
- Here is a screen recording with the issue:
![change cursor for invalid actions](https://user-images.githubusercontent.com/18304511/28575499-bfc49578-7159-11e7-8bce-6b41eab496b4.gif)

ericawright:
I thought we had an alert for this, similar to when you drop a folder? If not, we probably should

Shriram-Balaji:
I have some doubts on this, the issue requires that there is some sort of feedback when the user tries to upload an unsupported file, like a an alert of sorts? Please let me know if I am wrong. Thanks.

dannycoates:
I think what we'd like to do here is change the cursor to a [no-drop](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) when they're dragging, a folder, more than one file, or something other than a file.

Shriram-Balaji:
I have tried to solve the issue to some extent. Please review the recording 
![recording](https://user-images.githubusercontent.com/11358903/29212217-ee2eb8a8-7eba-11e7-834a-854985976819.gif)

dannycoates:
We won't fix this. We can't reliably change the cursor when dragging from off the page.

PR #505, opened 2017-08-11
Changed Cursor to No-Drop Ø symbol when trying to upload folders / multiple files

Shriram-Balaji:
#306

### DIFF
--- a/frontend/src/upload.js
+++ b/frontend/src/upload.js
@@ -46,6 +46,18 @@ $(() => {
 
       $link.attr('disabled', false);
 
+      const getCurrentTarget =  function(e) {
+        if (e.toElement) {
+            return e.toElement;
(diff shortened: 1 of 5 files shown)


frontend/src/upload.js
```diff
@@ -89,10 +101,11 @@ $(() => {
       });
 
       $uploadWindow
-        .on('dragover', () => {
+
+        .on('dragover', function(event) {
           $uploadWindow.addClass('ondrag');
         })
-        .on('dragleave', () => {
+        .on('dragleave', (event) => {
```
Johann-S: `event` is useless because you don't use it

Shriram-Balaji: Sorry, Will fix it and make a PR

Issue #446, opened 2017-08-04
Build a better 404 page for malformed URLs


pdehaan:
I was trying to copy/paste a URL to myself and it went horribly wrong:
https://send.firefox.com/download/f392

Basically you're treated to a very terse 404 page that just says "Not Found" in plain text.
We should probably make some fancier, styled "404" experience with headers and footers.

<img width="811" alt="mozilla_firefox" src="https://user-images.githubusercontent.com/557895/28988906-11feb8d0-7927-11e7-96ff-0645072d07d4.png">

Ref: [/server/server.js:139](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L139)

Not sure if we can do something like this instead:

```js
res.status(404).render('notfound');
```

Or if we need a `res.status(404).render('404')` view which informs the user that the URL was just horribly malformed and not downloaded/deleted/expired.

pdehaan:
In fact, here's a list of all the `res.sendStatus(404)` calls:

```
$ git grep -in "res.sendStatus(404);"
```

1. [server/server.js:124](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L124):    `res.sendStatus(404);` => `app.get('/exists/:id', ...)`
1. [server/server.js:132](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L132):    `res.sendStatus(404);` => `app.get('/exists/:id', ...)`
1. [server/server.js:139](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L139):    `res.sendStatus(404);` => `app.get('/download/:id', ...)`
1. [server/server.js:161](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L161):    `res.sendStatus(404);` => `app.get('/assets/download/:id', ...)`
1. [server/server.js:186](https://github.com/mozilla/send/blob/86fbd2a4b5e5de0b2d17e2c1c1100d5918e487e4/server/server.js#L186):    `res.sendStatus(404);` => `app.get('/assets/download/:id',  ...)`
[...]

PR #506, opened 2017-08-11
404 page

varghesethomase:
Fixing #446  by replacing all `res.sendStatus(404);` with `res.status(404).render('notfound');`

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -148,7 +148,7 @@ app.get('/exists/:id', async (req, res) => {
 app.get('/download/:id', async (req, res) => {
   const id = req.params.id;
   if (!validateID(id)) {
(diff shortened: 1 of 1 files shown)


abhinadduri:
@varghesethomase We should only render the notfound page for HTML endpoints; some of the 404s are API endpoints and will not be expecting a notfound HTML page.

varghesethomase:
@abhinadduri  Could you help me with identifying the APIs so that i can revert the necessary code?

varghesethomase:
I guess the only non XHR request out there is for the /download url. I am reverting all the other code.  And have updated the PR.

dannycoates:
Thank you @varghesethomase 😄

PR #507, opened 2017-08-11
Changed cursor to No-Drop Ø Symbol when trying to upload many files/f…

Shriram-Balaji:
#306 - Attempt to solve the specified issue.


dannycoates:
Thanks for the effort @Shriram-Balaji. This doesn't quite do what I think we want. I'm not sure its even possible to changed the cursor **while dragging a file from off page**. If that's the case we should close #306 as "won't fix"

PR #510, opened 2017-08-11
added precommit hook for format

dannycoates:
This one's for @pdehaan

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,11 @@
     "eslint-plugin-node": "^5.1.1",
     "eslint-plugin-security": "^1.4.0",
     "git-rev-sync": "^1.9.1",
+    "husky": "^0.14.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -118,5 +121,15 @@
...
     "test--browser": "browserify test/frontend/frontend.bundle.js -o test/frontend/bundle.js -d && node test/frontend/driver.js"
+  },
+  "lint-staged": {
+    "*.js": [
+      "prettier --single-quote --write",
+      "eslint",
+      "git add"
+    ],
+    "*.css": [
+      "stylelint"
```
pdehaan: nit: Not sure if we want to run `prettier --single-quote --write` here too or not. If so, we'd probably also need a `git add` step in case **prettier** rewrites the CSS.

PR #513, opened 2017-08-14
use svg for progress

dannycoates:
This trims about 22kB from the bundles

### DIFF
--- a/views/download.handlebars
+++ b/views/download.handlebars
@@ -20,8 +20,11 @@
   <div id="download-progress" hidden="true">
     <div id="dl-title" class="title"></div>
     <div class="description" data-l10n-id="downloadingPageMessage"></div>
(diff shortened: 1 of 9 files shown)


views/download.handlebars
```diff
@@ -20,8 +20,11 @@
   <div id="download-progress" hidden="true">
     <div id="dl-title" class="title"></div>
     <div class="description" data-l10n-id="downloadingPageMessage"></div>
-    <!-- progress bar here -->
-    <div class="progress-bar" id="dl-progress">
+    <div class="progress-bar">
+      <svg id="progress" width="166" height="166" viewPort="0 0 166 166" version="1.1">
```
dannycoates: This could be factored into another template but I'm considering a different template approach in general atm.

PR #514, opened 2017-08-14
use async and removed jquery from download.js

dannycoates:


frontend/src/download.js
```diff
@@ -41,76 +44,72 @@ function download() {
...
-        .formatValues('downloadNotification', 'downloadFinish')
-        .then(translated => {
-          notify(translated[0]);
-          $title.text(translated[1]);
-        });
+    links.setOpenInNewTab(false);
+    storage.totalDownloads += 1;
+    metrics.completedDownload({ size, time, speed });
+    progress.setText(' ');
+    document.l10n
```
pdehaan: These are promises too. Do we wait to `await` them, or ¯\\\_(ツ)\_/¯?

dannycoates: I thought about it, but it's fine (better?) that they happen async.

PR #527, opened 2017-08-20
attempting at improving readme rethorics

tmm88:
**with this PR I do basically aim at:**

- *attempting at improving readme rethorics;*

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ## What it does
 
-A file sharing experiment which allows you to send encrypted files to other users.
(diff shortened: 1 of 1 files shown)


ericawright:
Thank you for the PR, But, I disagree with both of these changes. The word "basically" is hedging, and doesn't actually make it more clear what _send_ does. 
"Via" also doesn't seem to add anything to this readme, and since it is used in the same sentence just a few words before seems repetitive.

Closing, please reopen if others disagree.

PR #533, opened 2017-08-23
minor UI fixes

youwenliang:
- Change SF Pro Display to SF Pro Text
- Better align percentage number
- Align copy text with input field
- Make input field url text smaller
- Fix footer social icon bug


dannycoates:
👍

Issue #558, opened 2017-09-08
Home screen icon is blurry


adamopenweb:
Minor issue. When adding to home screen on Android, the icon is kind of blurry.

STR:
1. Open Firefox for Android
2. Navigate to https://send.firefox.com
3. Go to the Firefox Menu, then Page, then Add to Home Screen

Expected Result:
Icon is high res

Actual Result:
Icon is low res (blurry)

(Loving this experiment btw)

ehuggett:
Its not as sharp as it could be for the "tile size" of my home screen, the edges also appear slightly clipped?

send.firefox.com offers two icon files, a 16x16px image with a white background and a 32x32px png with a transparent background. I assume we're both ended up with the transparent logo on our home screens? (I used a Firefox nightly)

note: the 16x16px favicon was removed in 53e8229 and is not available from send.dev.mozaws.net

sevaan:
It looks like the actual asset is blurry, probably due to it's size: https://send.firefox.com/favicon-32x32.48de2a43.png

According to [some research](http://www.emergeinteractive.com/insights/detail/The-Essentials-of-FavIcons-in-2017), a 32x32 icon just isn't good enough for the purposes of desktop/homescreen shortcuts. It turns out you need 13 different sizes to cover most cases.

<img width="752" alt="image" src="https://user-images.githubusercontent.com/473531/30446762-144ea964-9958-11e7-99d2-6ecc678caf36.png">

[You can find these icons sizes zipped up here.](https://github.com/mozilla/send/files/1304057/send.favicons.zip)

ehuggett:
Is send_logo.svg the source of ultimate truth for the logo? 🐵  

- According the to svg data the send logo should be `#3E3D40` but i see `#4f4e51` in favicon-228.png?
- The edges of the cloud appear a little clipped/flat, but my exports are the same, this seems to be as simple as we can't show a nice curve at these resolutions. Its a real shame we can't just use the svg for the favicon http://caniuse.com/#feat=link-icon-svg

Dimensions
- This image (svg imported and scaled to 228x205, place centred on a 228x228 canvas by **GIMP**) shows both with favicon-228.png coloured to blue and send_logo.svg to green (overlaps are darker).
![send_logo_diff_multiply](https://user-images.githubusercontent.com/8090731/30459256-79b97b04-99a7-11e7-97c0-85c34b82132f.png)


- SVG file exported by **Inkscape** (canvas _not_ resized to 228x228, image is 228x205)
[...]

PR #571, opened 2017-09-15
Centre logo

ehuggett:
Related to #558

Centre the Send logo within the image


ericawright:
Can you explain a bit more what has changed in this? I read through the bug, but I don't fully understand what was decided upon

ehuggett:
Sure, thanks for looking at this (its just a nit for the current usage of the logo, a couple of pixels or so off!)

The only addition should be ```transform="translate(-0.231,0.11948695)``` (its all on one line, so the github diff is not very useful).

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform
https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Basic_Transformations#Translation (has a better example for translate)

I believe this results in `0.231` being subtracted from all of the X coordinates and `0.11948695` added to all of the Y coordinates that describe the shape of the send logo within the 30 by 27 grid used. (This Moves it to the left and down slightly, Centring it more precisely than it was before)

The difference this makes can be seen at higher resolutions such as the images in [this comment](https://github.com/mozilla/send/issues/558#issuecomment-329808836), there are no visual differences at lower resolutions (they seem to be pixel for pixel identical at 30x27, as best as I can tell with GIMP)

The transform/translate could be removed by rewriting all of the coordinates to their new values, but I didn't manage to do this at the time with Inkscape. I don't think it was deemed necessary, but if you want I can have a look into doing this

ericawright:
Thank you for explaining. I can't really detect the difference myself, but it's been approved by Sevaan and it hasn't broken anything, so it's good by me 👍

PR #573, opened 2017-09-18
Support sending text or files

jondb:
Looking for ways to send secure messages. Could be a password or a private RSA certificate. The downside to saving it as a text file, uploading to send, and downloading remotely is keeping track of all of the artifacts. If I can generate a random string, copy, and paste it, there are no persistent file artifacts to accidentally leave behind.

I'd love to figure out a way to merge this into your primary branch. Looking for guidance.

### DIFF
--- a/app/fileManager.js
+++ b/app/fileManager.js
@@ -25,7 +25,7 @@ function convertPlainText(file) {
     const plaintextView = new TextDecoder().decode(file.plaintext);
     return plaintextView;
   }
-  return;
+  return null;
 }
 
 function openLinksInNewTab(links, should = true) {
--- a/app/fileManager.js
+++ b/app/fileManager.js
@@ -20,6 +20,14 @@ function saveFile(file) {
   URL.revokeObjectURL(downloadUrl);
 }
 
+function convertPlainText(file) {
+  if (file.type === 'text/plain') {
+    const plaintextView = new TextDecoder().decode(file.plaintext);
+    return plaintextView;
+  }
+  return;
+}
+
 function openLinksInNewTab(links, should = true) {
   links = links || Array.from(document.querySelectorAll('a:not([target])'));
   if (should) {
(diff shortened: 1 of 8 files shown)


app/fileManager.js
```diff
@@ -181,10 +189,17 @@ export default function(state, emitter) {
       const speed = size / (time / 1000);
       await delay(1000);
       await fadeOut('download-progress');
-      saveFile(f);
-      state.storage.totalDownloads += 1;
-      metrics.completedDownload({ size, time, speed });
-      emitter.emit('pushState', '/completed');
+      const pt = convertPlainText(f);
+      if (pt !== null) {
+        state.fileInfo.plaintext = convertPlainText(f);
```
pdehaan: Should this just be `state.fileInfo.plaintext = pt;` since we've already converted to plain text above on L192?

app/templates/download.js
```diff
@@ -1,28 +1,56 @@
 const html = require('choo/html');
 const progress = require('./progress');
 const { bytes } = require('../utils');
+const { fadeOut } = require('../utils');
```
pdehaan: Couldnt we just do a destructured import on the previous line, a la:

```js
const { bytes, fadeOut } = require('../utils');
```

app/templates/download.js
```diff
@@ -1,28 +1,56 @@
...
+            transfer.msg,
+            transfer.sizes
+          )}</div>
+        </div>
+      </div>
+    `;
+  } else {
+    div = html`
+      <div id="download-progress" class="fadeIn singlepane">
+        <div id="dl-title" class="title">Download Complete</div>
```
pdehaan: I _think_ this "Download Complete" needs a `${state.translate(...)}` wrapper of some sort for other locales.

app/templates/download.js
```diff
@@ -1,28 +1,56 @@
...
+        <div class="description">
+          <textarea class=pt-textarea>${state.fileInfo.plaintext}</textarea>
```
pdehaan: Nit: add double quotes around the `class` attribute value: `class="pt-textarea"`

jondb:
@pdehaan @dannycoates what are the next steps to merge this PR?

dannycoates:
@jondb sorry for the delay on this. I think something like this could be useful, but I think we should get some product and design feedback from @johngruen and @sevaan first.

app/fileManager.js
```diff
@@ -20,6 +20,14 @@ function saveFile(file) {
   URL.revokeObjectURL(downloadUrl);
 }
```
Maxvsg: [PACKET_2017-10-13_00-00-00.log](https://github.com/mozilla/send/files/1545234/PACKET_2017-10-13_00-00-00.log)
[PACKET_2017-10-26_06-00-00.log](https://github.com/mozilla/send/files/1545235/PACKET_2017-10-26_06-00-00.log)

dannycoates:
Thanks @jondb but it's unlikely we'll ever add this.

jondb:
We’ve built this feature and use it much more than files. 

Https://send.blend.com

fzzzy:
As of today, you can also paste text into send.firefox.com.

jondb:
How do I do this with my iPhone?
-- 
jondebonis@gmail.com

PR #574, opened 2017-09-21
Make upload button focusable (accessibility/tab navigation)

### LINKED ISSUE #563: Cannot access "Select file to upload button" using keyboard
The button to upload a file is not accessible via the keyboard...but should be!

ehuggett:
fixes #563

### DIFF
--- a/app/templates/welcome.js
+++ b/app/templates/welcome.js
@@ -23,11 +23,11 @@ module.exports = function(state, emit) {
         'uploadPageSizeMessage'
       )}</em></span>
       <form method="post" action="upload" enctype="multipart/form-data">
+        <input id="file-upload" type="file" name="fileUploaded" onchange=${upload} onfocus=${onfocus} onblur=${onblur} />
         <label for="file-upload" id="browse" class="${state.config
           .uploadButtonStyle}" title="${state.translate(
     'uploadPageBrowseButton1'
   )}">${state.translate('uploadPageBrowseButton1')}</label>
-        <input id="file-upload" type="file" name="fileUploaded" onchange=${upload} />
       </form>
     </div>
     ${fileList(state, emit)}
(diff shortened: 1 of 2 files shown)


assets/main.css
```diff
@@ -279,7 +279,19 @@ a {
 }
 
 input[type='file'] {
-  display: none;
+  width: 0.1px;
+  height: 0.1px;
```
ericawright: the width and height rules here don't seem necessary since it's `display: absolute` and `opacity: 0`

ehuggett: It doesn't seem important to me, but just in case, those rules were responsible for the 'hidden' form input fitting behind the label used as a button

browse_size

ehuggett:
> Oddly, once in a while the #browse element will show as selected, and others it simply won't. You can see this often by uploading one file then clicking upload another. I can't seem to figure out why though....perhaps it is in the wrong file?

I can reproduce this with Firefox ESR on Debian, it never works after clicking upload another file. I think this is because the javascript that adds the focus and blur events only runs on DOMContentLoaded.

app/templates/welcome.js
```diff
@@ -23,11 +23,11 @@ module.exports = function(state, emit) {
         'uploadPageSizeMessage'
       )}</em></span>
       <form method="post" action="upload" enctype="multipart/form-data">
+        <input id="file-upload" type="file" name="fileUploaded" onchange=${upload} onfocus=${onfocus} onblur=${onblur} />
```
ehuggett: The html attributes onfocus and onblur have the same issues as the css in Firefox, but it seems I can use them in the template as they get rewritten into javascript event listeners? (which do work in Firefox). I clearly need to read some documentation to understand this better, will it _always_ remove & replace the html onfocus attribute? (becuase if it doesn't, this will brake).

ericawright: It seems to be working for me now. 👍 
Can you re-base so this can merge smoothly?

ehuggett:
I think I just misused git, sorry.

I rebased, squashed and then force pushed so the reviewed/approved commits are no longer visible on github. 

I think i can put this back the way it was before (and then only rebase on master), should i?

ericawright:
@ehuggett yup, go ahead an put it back how it was, if you can, and I'll merge it

ericawright:
Actually, it looks correct now. I'll just test it again....

ehuggett:
I've got the original branch back (and rebased it, without squashing). I should be able to force push that if you haven't (re)started yet?

ericawright:
Nope, this is perfectly good as is. I prefer having a smaller commit history anyway.

PR #582, opened 2017-10-09
Add optional password to the download url

### LINKED ISSUE #69: Encrypt filename

### LINKED ISSUE #381: Add a pin/password for each file
Add the possibility to have a password/pin to protect each file.

dannycoates:
This adds an optional password to the download url of the shared file. Also, file metadata is now encrypted, so the server will no longer know the filename or type of all new files, not just password protected ones.

After a password is set the server requires verification of the password before returning both the encrypted file metadata and encrypted file.

Known issues:
- Does not work on Safari < 11, yet
- A few remaining css style issues
- With multiple browsers attempting to download a password protected file its possible for the valid password to be rejected if the server's nonce has changed. Re-entering the password with the new nonce will work as expected. I don't expect this to be common in real life.

Closes #381 
Closes #69

### DIFF
--- a/public/locales/en-US/send.ftl
+++ b/public/locales/en-US/send.ftl
@@ -34,6 +34,10 @@ sendAnotherFileLink = Send another file
 downloadAltText = Download
 downloadFileName = Download { $filename }
 downloadFileSize = ({ $size })
+unlockInputLabel = Enter Password
+unlockInputPlaceholder = Password
+unlockButtonLabel = Unlock
+downloadFileTitle = Download Encrypted File
 // Firefox Send is a brand name and should not be localized.
 downloadMessage = Your friend is sending you a file with Firefox Send, a service that allows you to share files with a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
 // Text and title used on the download link/button (indicates an action).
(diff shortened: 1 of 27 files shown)


public/locales/en-US/send.ftl
```diff
@@ -34,6 +34,10 @@ sendAnotherFileLink = Send another file
 downloadAltText = Download
 downloadFileName = Download { $filename }
 downloadFileSize = ({ $size })
+unlockInputLabel = Enter Password
+unlockInputPlaceholder = Password
+unlockButtonLabel = Unlock
+downloadFileTitle = Download Encypted File
```
flodolo: typo (encrypted)

public/locales/en-US/send.ftl
```diff
@@ -80,3 +84,6 @@ footerLinkAbout = About Test Pilot
 footerLinkPrivacy = Privacy
 footerLinkTerms = Terms
 footerLinkCookies = Cookies
+requirePassword = Require a password to download this file
```
flodolo: Can you use a string ID that describes better the role of this string, or add a localization comment to explain the context?

public/locales/en-US/send.ftl
```diff
@@ -80,3 +84,6 @@ footerLinkAbout = About Test Pilot
...
+addPassword = Add Password
```
flodolo: Same here, can you add some context? I assume it's a button, or a link.

Issue #594, opened 2017-10-13
The "Password" and "Copy to clipboard" fields are misaligned


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the browser and navigate to https://send.dev.mozaws.net.
2. Upload a file and click on the "Require a password to download this file" checkbox.
3. Observe the "Password" field position in comparison with "Copy to clipboard" field.

[Expected result]:
- The "Password" and "Copy to clipboard" fields have the same alignment.

[Actual result]:
- The "Password" and "Copy to clipboard" fields are misaligned. 

[Notes]:
- Here is a screen shot of the issue:
![aliniament](https://user-images.githubusercontent.com/31244235/31552712-ef7c6dce-b040-11e7-8654-0276e23c7c50.png)

sevaan:
The password box should be aligned to be underneath the checkbox. Also, it should be blue (same blue as the URL field outline) to show it as active.

<img width="653" alt="screen shot 2017-10-16 at 1 48 06 pm" src="https://user-images.githubusercontent.com/473531/31626697-bf2cb644-b278-11e7-8ddf-078c56693213.png">

Softvision-CristinaBadescu:
I can confirm that this issue is not reproducible anymore on Send dev server, using latest Nightly build 58.0a1 (2017-10-23) on Windows 10x64, Mac 10.12 and Ubuntu 14.04 x64.

Issue #587, opened 2017-10-11
Remaining password UI bits


dannycoates:
There's a few remaining things from the design spec that aren't done and some areas where I intentionally didn't follow the spec that should be reviewed. https://mozilla.invisionapp.com/share/CTDDQDXME#/screens/252367497

### Differences
- The 'add password' input box is plain text so the dots for the placeholder seemed to imply the text would be hidden or was already set to something. I changed it to 'Password'.
- I changed the label of the button from 'Submit' to 'Unlock'
- I took a few liberties with the strings

### Incomplete
- [x] responsive password input boxes
- [x] styled checkbox
- [x] "When the user clicks the "Require a password" checkbox, the link field is not disabled." --- it should be
- [x] "Try again?" should be "Try again."
- [x] Update checkbox style to match the preferences pane (see screenshot in comment below)

ericawright:
> I changed the label of the button from 'Submit' to 'Unlock'
Nice change!

Softvision-CristinaBadescu:
Hello,

I've went through the mock-ups in order to write test cases for password functionality and I've observed certain differences between the mock-ups and the experiment on dev server. For example:
- The "edit password" mode doesn't exists at the moment.
- When user checks box, the link field is not disabled.
- The title is different from the mock-ups: "The link to your file will expire after 1 download or in 24 hours." instead of "This encrypted link will expire after 1 download or in 24 hours".
- When typing a wrong password, the text has another punctuation: "Try again?" instead of "Try again:".

I strongly recommend using a hidden password instead of a plain text. Also, the user should have a "show/hide" option for the password when typing it; in this way some security issues can be avoided.
 
Are we going to implement these on the next release? If yes, I can log them so it will be easier to track.

wresuolc:
Thanks for the feedback.  Replying to your bullet points:

- We're not planning on putting an 'edit password' mode in at this point
- This sounds like a bug, thanks for finding it.  I'll add that to the original post
- We think the one we have already translated is ok
- Thanks for finding this too.  I'll update that in the original post as well.

The password is hidden when entering the password when downloading a file, but we don't think it's necessary when creating the password and keeps the UI simpler (eg. don't have to enter twice).  Sevaan did some research on this.

Softvision-CristinaBadescu:
Hello,

Thanks for your reply. Also, do you think I should file separately the issues in order to track them or is it sufficient that you included them in the original post?

youwenliang:
[...]

PR #605, opened 2017-10-20
align the "Password" and "Copy to clipboard" fields.

ericawright:
the result of this css change: 

screen shot 2017-10-20 at 4 22 34 pm

Fixes: #594.

### DIFF
--- a/assets/main.css
+++ b/assets/main.css
@@ -652,7 +652,7 @@ tbody {
   display: flex;
   flex-wrap: nowrap;
   width: 80%;
-  padding: 10px 20px;
+  padding: 10px 5px;
 }
 
 /*    upload-error    */


sevaan:
Alignment of boxes looks good. Noticed two other issues though (let me know if I should file separate issues, but...):

1: Text in fields should be middle-aligned:

image

2: The checkbox should be bigger: 16x16

ericawright:
@sevaan yes, this one is only for the alignment. the checkbox, I thought was [supposed to be](https://github.com/mozilla/send/issues/587#issuecomment-338167607) 20x20? I'm currently working on the checkbox style for a different PR.

sevaan:
The sketch file has it at 16x16, but let's see it at 20x20 anyway and decide then! I'll file a separate issue for the field labels then.

ericawright:
also, in your spec, the password input and checkbox is slightly (5px) indented, but in the other one ^ is exactly aligned - which is correct?

sevaan:
> in the other one ^ is exactly aligned

Which other one are you referring to? The box and field should be slightly indented. It's like the top field is the parent and this is a sub-option rather than a separate thing.

ericawright:
@sevaan the exported spec in the above [referenced comment](https://github.com/mozilla/send/issues/587#issuecomment-338167607) has no indent.

Issue #606, opened 2017-10-20
Disable copying link when password not completed


ericawright:
from #587.

>"When the user clicks the "Require a password" checkbox, the link field is not disabled." --- it should be

And mocks from https://github.com/mozilla/send/issues/587#issuecomment-338167607

Issue #600, opened 2017-10-17
[Enhancement] Unchecking the require password checkbox should also clear the typed password


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Open the browser and upload a file to https://send.dev.mozaws.net.

[Steps to reproduce]:
1. Check the require password checkbox and type something in the "Password" field.
2. Uncheck the require password checkbox.
3. Check again the require password checkbox.
4. Observe the "Password" field.

[Expected result]:
- The "Password" field should be cleared and only the placeholder will be displayed after checking again the checkbox.

[Actual result]:
- The "Password" field contains the previously typed password.

[Notes]:
- Here is a screen recording of the actual behavior:
![enhancement send](https://user-images.githubusercontent.com/31244235/31668926-322c15de-b35c-11e7-8609-d8af4cfbaf06.gif)

Softvision-CristinaBadescu:
I can confirm this issue is no longer reproducible on Send dev server, on latest Nightly build 58.0a1 (2017-10/30) using Windows 10 x54, Mac 10.12 and Ubuntu 14.04 x64.

PR #608, opened 2017-10-23
disable copying link when password not completed

ericawright:
select require password, the link and "copy" button get half opacity, and the button is disabled. When the checkbox is deselected or password is submitted the opacity is restored, and button is enabled.

result: 

screen shot 2017-10-23 at 9 55 59 am

Fixes: #606.
Fixes: #600.
Related: #587.

### DIFF
--- a/app/templates/share.js
+++ b/app/templates/share.js
@@ -64,7 +64,9 @@ module.exports = function(state, emit) {
       );
       await delay(2000);
       input.disabled = false;
-      copyBtn.disabled = false;
+      if (!copyBtn.parentNode.classList.contains('wait-password')) {
+        copyBtn.disabled = false;
+      }
       copyBtn.classList.remove('success');
       copyBtn.textContent = state.translate('copyUrlFormButton');
     }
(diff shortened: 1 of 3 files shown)


ericawright:
@clouserw can you find me a re viewer for this please?

wresuolc:
@fzzzy has volunteered.  thanks!

fzzzy:
This looks good. I'd like to figure out how to build and test this locally so I can just verify it does what I think it does.

fzzzy:
This works well, but I did notice one tiny bug. I think it is mostly a nit.

To reproduce, click the copy button, then click the password button while the green checkmark is still visible in the copy button. The copy button will be rendered as faded out, but it will not be disabled and the user will be able to click it to copy the url again.

I think this is because after the `await delay(2000)`, `copyBtn.disabled = false;` is being executed even if the password checkbox is now checked. This is a weird corner case and it is probably fine if it doesn't get fixed.

ericawright:
nit has been fixed, that @fzzzy

Issue #610, opened 2017-10-23
P in "Add Passwords" should be lowercase


sevaan:
The P in the Add Passwords button should be lowercase, to match the casing of the button label above it:

![image](https://user-images.githubusercontent.com/473531/31898623-7f2b852e-b7e7-11e7-8424-f3a9c9bfbf8c.png)

Issue #599, opened 2017-10-17
Password isn't wrapped if it is very long


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the browser and navigate to https://send.dev.mozaws.net.
2. Upload a file and click on the "Require a password to download this file" checkbox.
3. Type in the "Password" field an excessive number of characters (at least 75 characters) and click on "Add password" button.
4. Observe how the password is displayed.

[Expected result]:
- The password is wrapped.

[Actual result]:
- The password is displayed on a single row.

[Notes]:
- If the password has more that 150 characters, a horizontal scrollbar appears and the password is not wrapped to a second row.
- Here is a screen shot of the issue:
![very long password](https://user-images.githubusercontent.com/31244235/31664968-15832d3e-b350-11e7-9711-7130644abd37.png)

wresuolc:
Lets cap passwords at 64 characters.

Softvision-CristinaBadescu:
I can confirm that this issue is no longer reproducible and the password has a 64 characters limit, on Send dev server, on latest Nightly build 58.0a1 (2017-10-30) using Windows 10 x64, Mac 10.12 and Ubuntu 14.04 x64.

Issue #591, opened 2017-10-12
The added password is wrongly displayed when there are more than two consecutive spaces in it


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the browser and navigate to https://send.dev.mozaws.net.
2. Upload a file and select the "Require a password to download this file" checkbox.
3. Add a password that contains at least two consecutive spaces and observe the displayed password.

[Expected result]:
- All password characters are correctly displayed.

[Actual result]:
- The password is displayed incomplete: only one space is shown.

[Notes]:
- The password should be fully displayed or maybe the space character could be a forbidden character for the password.
- Here is a screen recording of this issue:

![send spaces](https://user-images.githubusercontent.com/31244235/31506997-cf94d09c-af81-11e7-978b-c3a6a6e233b4.gif)

Softvision-CristinaBadescu:
I can confirm that this issue is no longer reproducible on Send dev server, on latest Nightly build 58.0a1 (2017-10-30) using Windows 10 x64, Mac 10.12 and Ubuntu 14.04 x64.

PR #612, opened 2017-10-23
Password UI nits

ericawright:
smaller screens:
screen shot 2017-10-24 at 12 50 34 pm
screen shot 2017-10-24 at 12 50 07 pm

Tested on release versions of Chrome and Firefox

- Turns out HTML just doesn't remember spaces - so need to convert to `&nbsp;` to display passwords with spaces properly.

Fixes: #587.
Fixes: #610.
Fixes: #599.
Fixes: #591.
Fixes: #613.


public/locales/en-US/send.ftl
```diff
@@ -85,5 +85,5 @@ footerLinkPrivacy = Privacy
 footerLinkTerms = Terms
 footerLinkCookies = Cookies
 requirePasswordCheckbox = Require a password to download this file
-addPasswordButton = Add Password
-incorrectPassword = Incorrect password. Try again?
+addPasswordButton = Add password
```
flodolo: It's OK to change case without using a new ID.

public/locales/en-US/send.ftl
```diff
@@ -85,5 +85,5 @@ footerLinkPrivacy = Privacy
...
+incorrectPassword = Incorrect password. Try again.
```
flodolo: Unfortunately [this needs a new string ID](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings) if we want locales to update accordingly.

sevaan:
Let's copy the formatting of the URL field, with the field on top and the button below.

public/locales/en-US/send.ftl
ericawright: thank you!

ericawright:
@clouserw can you find a reviewer for this too?

dannycoates:
This looks great, thanks @ericawright! 👍

PR #624, opened 2017-10-31
set a default MIME type in file metadata

### LINKED ISSUE #620: Set a default MIME type in metadata?
I'm not aware of any problems this is causing (noticed by chance), but if I upload a file that does not have an extension then the "type" field in the (encrypted) metadata is empty.

Server version: 5211339
Browser: Firefox 52.4.0 (64-bit) on Debian

https://github.com/mozilla/send/blob/52113395db7ffd8c2b25804908743d68bfe8f0e5/app/fileSender.js#L195-L197

I added `|| 'application/octet-stream'` to L197 in https://github.com/ehuggett/send/commit/58555a6b85f90d8fbff0781d6bda9b7d43cbe906, and when uploading files with names that do not have an extension this "default" MIME type was used instead

ehuggett:
fixes #620 

Summary: file.type can be empty (such as when the filename has no extension) , provide a default value to use should this happen when collecting the file metadata

### DIFF
--- a/app/fileSender.js
+++ b/app/fileSender.js
@@ -194,7 +194,7 @@ export default class FileSender extends Nanobus {
         JSON.stringify({
           iv: arrayToB64(this.iv),
           name: this.file.name,
-          type: this.file.type
+          type: this.file.type || 'application/octet-stream'
         })
       )
     );


dannycoates:
Thank you @ehuggett 😄

Issue #622, opened 2017-10-31
The password preview should be bold


SoftVision-CosminMuntean:
[Notes]:
- Based on [mock-ups](https://mozilla.invisionapp.com/share/CTDDQDXME#/screens/252367497) the password preview should be bold.

**[Affected versions]:**
- Firefox 56 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Steps to reproduce]:**
1. Open the browser and navigate to https://dev.send.mozaws.net. 
2. Upload a file and select "Require a password to download this file" checkbox.
3. Type a password and click the "Add password" button.
4. Observe the password preview.

**[Expected result]:**
- The password preview should be bold.  

**[Actual result]:**
- The password preview is not bold.

[Additional Notes]:
- Here is a screenshot with the issue:
![screenshot-2017-10-31 firefox send](https://user-images.githubusercontent.com/19226534/32225860-84dfeb54-be4f-11e7-8e56-bb95e2e73ea6.png)

- Here is a screenshot with password preview from mock-ups:
![b51c0dcb-a408-4975-8bb8-86c22a07a7da](https://user-images.githubusercontent.com/19226534/32225687-d3b5c16e-be4e-11e7-8d87-6e8447601bbe.png)

Issue #621, opened 2017-10-31
The "Password:" string is no longer displayed after adding a password


Softvision-CristinaBadescu:
**[Affected versions]:**
- Firefox 56 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Steps to reproduce]:**
1. Open the browser and navigate to https://dev.send.mozaws.net. 
2. Upload a file and select "Require a password to download this file" checkbox.
3. Type a password and click the "Add password" button.
4. Observe the password preview.

**[Expected result]:**
- The following text appears: "Password: [the password you typed]".

**[Actual result]:**
- The "Password:" string is missing and only the typed password is displayed.

**[Notes]:**
- Here is a screen recording of the issue:
![send missing password](https://user-images.githubusercontent.com/31244235/32223728-34db6c80-be47-11e7-84a3-7e12028bd91d.gif)

Softvision-CristinaBadescu:
I can confirm that this issue is no longer reproducible using latest Nightly build 58.0a1 (2017-11-06) on Windows 10 x64, Mac 10.12 and Ubuntu 14.04 x64.

PR #625, opened 2017-10-31
Password string

ericawright:
Fixes: #621.
Fixes: #622.

### DIFF
--- a/public/locales/en-US/send.ftl
+++ b/public/locales/en-US/send.ftl
@@ -87,3 +87,5 @@ footerLinkCookies = Cookies
 requirePasswordCheckbox = Require a password to download this file
 addPasswordButton = Add password
 passwordTryAgain = Incorrect password. Try again.
+// This label is followed by the password needed to download a file
+passwordResult = Password required to download this file:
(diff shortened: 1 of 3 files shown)


public/locales/en-US/send.ftl
```diff
@@ -87,3 +87,4 @@ footerLinkCookies = Cookies
 requirePasswordCheckbox = Require a password to download this file
 addPasswordButton = Add password
 passwordTryAgain = Incorrect password. Try again.
+passwordResult = Password:
```
flodolo: Can you add a short localization note? e.g.

```
// This label is followed by the password set to download a file
```

Is it just me or #622 shows a completely different string?

ericawright: It does, though I believe the consensus was to change it to just be "password", though it's possible that's just what it became in the last iteration....

ehuggett: https://github.com/mozilla/send/issues/381#issuecomment-327930546 has the mockups (last image in "sender" shows the string "Password required to download this file:")

ericawright: I double-checked with some folks, and the long one is preferred. Changing it, thanks!

dannycoates:
This looks good, but I made some other changes in another PR, so I'm closing this...

To maybe make RTL better, in #632, I made the password an ftl variable, and also some inside baseball changes to make other weird passwords (besides multiple spaces) work better.

Issue #623, opened 2017-10-31
The "Password" field isn't displayed after restarting the browser


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the browser and navigate to https://send.dev.mozaws.net.
2. Upload a file and click on the "Require a password to download this file" checkbox.
3. Restart the browser using the Developer Toolbar and observe the "Password" field.

[Expected result]:
- The "Password" field is displayed.

[Actual result]:
- The "Password" field isn't displayed and the require password checkbox is checked.

[Notes]:
- After restarting the browser if you uncheck the "Require password" checkbox, the "Password" field appears.
- This issue is reproducible when restarting the browser with the "Show your windows and tabs from last time" preference enabled.
- This issue is also reproducible using the "Restart to update Nightly" option. 
- Here is a screen recording of the issue:
![restart password field](https://user-images.githubusercontent.com/31244235/32227259-8c1aafc2-be53-11e7-9108-a9eeb3af2706.gif)

ehuggett:
I'm not going to attempt the expected action of 'The "Password" field is displayed.', but I will open a PR as this partially addresses the issue

Two changes https://github.com/ehuggett/send/commits/623-password-field
1. `togglePasswordInput()` was setting the state of the "copy" button based on the state of the checkbox but would always invert the visibility of the password form. Having both set by the state of the checkbox allows it to "get back in sync" / "catch up" with the rest of the form should its parts ever be mismatched
2. autocomplete="off" on the checkbox prevents firefox from persisting its state across page loads, when the browser restarts it is unchecked (and the password input hidden, both ignoring any previous state)

Further information from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete

> The autocomplete attribute also controls whether Firefox will, unlike other browsers, [persist the dynamic disabled state and (if applicable) dynamic checkedness](https://stackoverflow.com/questions/5985839/bug-with-firefox-disabled-attribute-of-input-not-resetting-when-refreshing) of an <input> across page loads. The persistence feature is enabled by default. Setting the value of the autocomplete attribute to off disables this feature. This works even when the autocomplete attribute would normally not apply to the <input> by virtue of its type. See [bug 654072](https://bugzilla.mozilla.org/show_bug.cgi?id=654072).

dannycoates:
[...]

PR #626, opened 2017-11-02
Partial fix for #623

ehuggett:
please see my comment https://github.com/mozilla/send/issues/623#issuecomment-341301803 for a description.

Does not result in the expected action, but allows the password form to work correctly if Firefox restarts while it is open


dannycoates:
Thank you, @ehuggett. This is a good solution. I don't think it's important to persist the checkbox checked state between browser restarts.

Issue #609, opened 2017-10-23
Text in fields should be middle-aligned


sevaan:
The text in the URL and the password fields are slightly higher than the text in the attached labels:

![image](https://user-images.githubusercontent.com/473531/31898474-49c7b7d6-b7e7-11e7-98b0-80915d065860.png)

They should all be middle-aligned.

tulikavijay:
I would like to work on this.

tulikavijay:
I am new to open source so please guide me a bit. I want to ask, Where should I make changes?

ericawright:
Hi @tulikavijay, Thank you!
When you open the project you can find the styling in `/assets/main.css`, this can likely be fixed with a few css changes. Once you have it committed, push it up to your fork of the project and make a pull request to us.
You can ping me or @dannycoates if you get stuck.

SoftVision-CiprianMuresan:
@sevaan, the text in the box and the text in the label are now aligned with each other, but the text from the label was moved up a bit instead of moving the text from the box down.
Is this a suitable solution?

sevaan:
@SoftVision-CiprianMuresan No, it should be the opposite: The label is too high, and the button is correct.

anavgupta:
Hi @sevaan, I am looking to contribute to my first Open Source Project. Could you please point out the issue after the previous merge so that I can try to fix it.

sevaan:
@anavgupta Great! Welcome aboard and thanks for contributing. Could you provide a screenshot of the issue after the previous merge (I'm not set up to build at the moment), and I'll give you the details on the fix.

shikhar-scs:
@sevaan If this issue ( bringing the fields down ) isnt fixed till now , then may I do that and file a pr?

bmintz:
@shikhar-scs of course!

shikhar-scs:
Actually I'm stuck upon another issue in mozilla send so I wouldn't mind @himanish-star taking over this issue

himanish-star:
@bmintz , I'll take over this issue

PR #649, opened 2017-11-22
#609 share interface: align text in input and button

ovlb:
Bug ([#609](https://github.com/mozilla/send/issues/609)): Text in input and button are not aligned on the middle axis

Fix: Add `padding-bottom` on the button elements to push their contents a bit higher
Send’s share interface with middle aligned content in inputs and button


dannycoates:
Thank you @ovlb 😄

Issue #634, opened 2017-11-06
"Add Password" button looks wrong on mobile safari


dannycoates:
![img_0275](https://user-images.githubusercontent.com/87619/32456405-54a0eb2a-c2da-11e7-946a-72ea5ce129ca.PNG)

ovlb:
Hej,
this also happens in Firefox 10.2, Chrome 62 and every other browser I just tested. My iOS version is 11.1.1. 
The easy fix seems to be changing `<input type="submit" value="Add password" …>` to `<button type="submit" …>Add password</button>`. Is this acceptable? If so, I would open a pull request.

Screenshot Firefox
![img_1924](https://user-images.githubusercontent.com/7645967/33166639-3c8e8b8c-d03b-11e7-91a5-b08bf7417a99.png "Incorrectly rendered «Add password» button in Firefox on iOS")

Screenshot Chrome
![img_1925](https://user-images.githubusercontent.com/7645967/33166664-5a36fe94-d03b-11e7-8a81-ab1a3e1e391e.png "Incorrectly rendered «Add password» button in Chrome on iOS")

youwenliang:
Should probably also add "-webkit-appearance: none" to the button class.

ovlb:
In Safari this also fixes the issue, making it superfluous to change the markup. I would create a fix for this. Then it should probably be merged in a dev branch to test it across browsers? Or is your Webpack config (I just can’t get my head around Webpack) creating a local dev server which I can access on my iPhone to test it? In the moment I just have the Safari remote debugger available.

ovlb:
I’ve tested `-webkit-appearance: none` against Firefox, Chrome, Safari, Opera and the Apple WebViews and it fixes the layout issue. Thanks, @youwenliang. Would open a PR later on.
Results can be seen in this CodePen: https://codepen.io/ovl/full/qVYgoO/

dannycoates:
fixed

PR #650, opened 2017-11-23
#634: overwrite appearance of password submit input

ovlb:
Bug: The button to set a password had a different look than other buttons on the site

Fix: add `-webkit-appearance: none` on the input element to force browser to discard the default styling. For a reduced test case demonstrating the fix see https://codepen.io/ovl/full/qVYgoO/


dannycoates:
Thank you @ovlb 😄

Issue #666, opened 2017-12-08
When hosting http:// based instance, download links are https:// (scheme confusion)


claudijd:
I tried deploying and instance on the bench to play around with send and noticed that when the docker instance is hosting content over http:// and I drag and drop a link, the download URL that's offered is https:// on the same http:// hosting port, resulting in a bad link.

Now, if send is running behind an ssl termination proxy, this is all well and good, but if you are testing locally, this presents a mis-match.  I'm wondering if the served download URL on an http-based instance should be http:// and then let whatever front end SSL/TLS termination proxy do the translation.

Here's a pic example of what I'm seeing in case it's not clear:

<img width="1078" alt="screen shot 2017-12-08 at 3 10 44 pm" src="https://user-images.githubusercontent.com/794283/33783251-fd4736ea-dc29-11e7-88c4-7748c4893552.png">

Note: http:// transport, but https:// download URL.

dannycoates:
This is sort of a "feature" of using the `NODE_ENV=production` environment variable in that it forces the generated link to be https. We should probably change `docker-compose.yml` to not set `NODE_ENV`

claudijd:
Yeah, I can confirm, removing that and rebuilding and docker-compose up that that addresses the issue.  Happy to send a PR for the docker-compose.yml (assuming prod deployment over-write this env var)

PR #667, opened 2017-12-08
Make develop the default NODE_ENV

claudijd:
This fixes https://github.com/mozilla/send/issues/666 by not assuming production in the docker-compose.yml.

**Important Note:** before deploying this, please make sure the production infrastructure actually sets the NODE_ENV variable as this could pose a production => development node env regression if not.


dannycoates:
Thanks @claudijd! Production doesn't use `docker-compose.yml` so we're good.

PR #673, opened 2017-12-18
Mask password on upload

claudijd:
When a user enters a password for a given file download, mask it so it doesn't show to someone looking over their shoulder.  I also briefly looked into how to mask the echo'd password, which comes out in pre tags, but would love some advice on how to approach that.  Advice is welcomed!

### DIFF
--- a/app/templates/uploadPassword.js
+++ b/app/templates/uploadPassword.js
@@ -13,6 +13,7 @@ module.exports = function(state, emit) {
     </div>
     <form class="setPassword hidden" onsubmit=${setPassword} data-no-csrf>
       <input id="unlock-input"
+        type="password"
         class="unlock-input input-no-btn"
         maxlength="64"
         autocomplete="off"


claudijd:
Behavior change this is attempting to achieve is this...

BEFORE

screen shot 2017-12-18 at 3 27 29 pm

AFTER

screen shot 2017-12-18 at 3 27 46 pm

Basically, make the password hidden while typing it in.

dannycoates:
After the password is set, we show a set state under the link. Currently, it includes the plain text password. I would be surprised if my hidden input was revealed after I click the button. So, if we want the input to be hidden, I think the display of the set state should be obscured by default as well but include an option to reveal it.

himanish-star:
@dannycoates , could I make these modifications or @claudijd are you still working on it?

claudijd:
@himanish-star please, feel free to make the mods, I was just trying to help, but my JS isn't that strong.

himanish-star:
@claudijd , Thanks but shikhar has already filed the PR and I guess he's done it right

claudijd:
Closing in favor of #686

PR #680, opened 2017-12-21
adjusted line height of label : fixes #609

### LINKED ISSUE #609: Text in fields should be middle-aligned

himanish-star:
This should fix #609 

Preview of changes

image


dannycoates:
looks good to me. thanks @himanish-star 😄

Issue #661, opened 2017-12-08
[Enhancement] Show number of remaining downloads


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 57.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Start the Firefox browser with and navigate to "https://send.stage.mozaws.net/" page.
2. Upload a file and set it to expire after 20 downloads.
3. Navigate to the homepage and observe the newly uploaded file in the file info table.

[Expected result]:
- The number of remaining downloads is displayed in the table.

[Actual results]:
- No such information is displayed.

[Notes]:
- This information could also be displayed on the file's download page.

shikhar-scs:
@SoftVision-CiprianMuresan I am working on this. Though I had a doubt.
Do we need the information of `after downloads` to be dynamic ( i.e. that it should decrease after every download ) or not.

sevaan:
> Do we need the information of after downloads to be dynamic ( i.e. should it decrease after every download ) or not.

Yes, that's correct. It's a running tally of how many downloads are left.

On the home page, please let's change the "Expires In" header to "Time" and add a column to the right with the header "Downloads". The number in the download column should be center aligned.

<img width="649" alt="image" src="https://user-images.githubusercontent.com/473531/34275226-521e2ad0-e66a-11e7-99e3-0fafe38db900.png">

I feel like those last columns could be crushed to the right so that more of the file name shows.

shikhar-scs:
@sevaan I've made the required changes and also have filed a PR #681 . Please review.

sevaan:
I don't do code reviews, I'm afraid. My programming skills are lacking. @dannycoates do you mind taking a look?

shikhar-scs:
@dannycoates Yes Please

PR #681, opened 2017-12-21
Show Download Limits on upload page: Solves #661

shikhar-scs:
As suggested by @SoftVision-CiprianMuresan at #661, the downloads available can now be seen on the main upload page. This is the basic work done. 

Preview 

screenshot from 2017-12-21 18-37-12

**Due to non-availability of translated texts, I could not change the table headers as to my convenience and so the `Expires in` column is repeated. Also an svg image icon can be used instead of the word d'load (complete downloads was not fitting in the table cell) However the download limits can now be seen on the main page.**

Please review @sevaan @dannycoates @SoftVision-CiprianMuresan


himanish-star:
@shikhar-scs ,cool

shikhar-scs:
@sevaan
 I've edited the column headers as reuired (even though something better would be if we could individually put `${state.translate(<table header name>)}` but that wont be much of a task I guess once we have the required statements in all the languages.
Also, the width of file name column has been improved considerably as required by you. Here's a preview.


screenshot from 2017-12-22 09-34-38


Please review

shikhar-scs:
@dannycoates mind giving a review??

sevaan:
Looks great, @shikhar-scs. Thanks!

shikhar-scs:
@sevaan welcome. Then merge ??

sevaan:
I don't do code reviews, so wait until @dannycoates  gives it a look over. But from a front-end visual POV, looks good.

shikhar-scs:
@sevaan thanx a lot for your review. I'll surely wait for @dannycoates

shikhar-scs:
@dannycoates mind reviewing here please ?

dannycoates:
This is going to need more work, including server api changes, in order to stay updated between page refreshes and as the download count changes. I think it'd be better if I take on this task. I'm sure @shikhar-scs  could do it, but there's some other aspects of the design I'd like to change that also affect this, and it'd probably take us just as long to talk about it as it would for me to just do it.

shikhar-scs:
@dannycoates 

> and it'd probably take us just as long to talk about it as it would for me to just do it.

Even if that's the case, still I would like to work on this as I had already invested much time and thus, this PR going away like this is not good for me. 

With time I've only understood more and more of the code base.

And I believe with a little bit of guidance I'll be able to do all of the stuff you require. If you could just make one long comment as to where the rest of the changes have to be made, I'll see to my self that they work. And if they dont I'll be more than happy to let you take over.

> I'm sure @shikhar-scs could do it, but there's some other aspects of the design I'd like to change

Once I'm done with the backend stuff you can incorporate all the design changes.(or mention it in the comment itself and I'll do that ) Rest is up to you.

shikhar-scs:
@dannycoates any thoughts on this ? :crossed_fingers:

dannycoates:
Ok @shikhar-scs. If you're in this for the long haul I'll help guide you through it. You can expect this to be a pretty big effort and will likely take a fair amount of time. I will have some time later this week to come up with a plan. Stay tuned.

shikhar-scs:
yep Sure I am ready to invest time if the need be and also tell me if I need to remind you about the same :smile:

dannycoates:
Ok, @shikhar-scs, here we go...

In `app/fileManager.js` there's a function named `checkFiles`. It currently calls `exists` which makes a request to the server to see if it still exists. We'll want to instead call a new function that makes a request to `/api/metadata` in order to get the download counts. Fortunately I think we can use the `getMetadata` method of `FileReceiver`. We'll also probably need to change the `rerender` logic a bit too.

There's a `setInterval` at the bottom of `fileManager.js` that rerenders periodically to keep the expire time up to date. Now we also want to keep the counts up to date so we should change that `render` call to `checkFiles`.

The rest should just be updating the UI parts, which you've already started on. In the downloads column I think we should show both the number of downloads that happened `dtotal` and the limit `dlimit` as (psuedocode) `"${dtotal} / ${dlimit}"` ex. "2 / 5"

shikhar-scs:
Awesome !!! I'll get started with it right away. Thank you.

himanish-star:
@shikhar-scs , awesome

shikhar-scs:
@dannycoates I'll tell you my approach.

> It currently calls exists which makes a request to the server to see if it still exists.

Using the else part of if(!ok), I wrote my checkDL method to receive metadata respective to that file. This ensured that we are asking metadata of files which actually exist.

> We'll want to instead call a new function that makes a request to /api/metadata in order to get the download counts. Fortunately I think we can use the getMetadata method of FileReceiver.

In the new method, taking cue from this following snippet on how to use the `FIleReceiver methods`, 

https://github.com/mozilla/send/blob/bace117ada440e14a40c22f070ae40a1b85ab803/app/fileManager.js#L188-L196

I wrote a similar method in order to use the getMetadata method of FileReceiver.
```
const url = `/api/metadata/${file.id}`;
const receiver = new FileReceiver(url, file);
```

However, the file which is sent in case of the `on('preview'` example to FileReceiver is taken from state.fileinfo (see permalink above ). But, the file which is used in my method is taken from state.storage.files. 

https://github.com/mozilla/send/blob/bace117ada440e14a40c22f070ae40a1b85ab803/app/fileManager.js#L64

The difference is that files taken from state.storage.files don't have the following or similar attributes namely `file.key` , `secretKeyPromise`, `authKeyPromise`, or `metaKeyPromise`. Because of this, when, `b64ToArray(file.key),` method is called on the new file.key, it returns a null error. And in general, every constructor function returns errors in the console. But, the ones from `state.fileInfo` do contain them and thus, this works smoothly in that case. 

I also tried using then `state.fileInfo` but, I guess it is a specific file related option, something not suitable to be used in case of checkFiles() where we operate collectively, on all the files.

>  There's a setInterval at the bottom of fileManager.js that rerenders periodically to keep the expire time up to date. Now we also want to keep the counts up to date so we should change that render call to checkFiles.
The rest should just be updating the UI parts, which you've already started on. In the downloads column I think we should show both the number of downloads that happened dtotal and the limit dlimit as (psuedocode) "${dtotal} / ${dlimit}" ex. "2 / 5"

This however looks rather simpler comparatively to manage.

Could you please guide on how to tackle the above problem.

dannycoates:
> Because of this, when, b64ToArray(file.key), method is called on the new file.key, it returns a null error.

I think you're close but I have a feeling you aren't up to date with the master branch.

shikhar-scs:
> I think you're close but I have a feeling you aren't up to date with the master branch.

True that, life saviour :smile: .

shikhar-scs:
Ok @dannycoates, I've implemented changes and please have a look.

I guess dynamic changes are still not occurring. They do occur on refresh though. Here's a preview.

dlimiitt

The main difference between dynamic display of time and downloads is that file.createdAt is a const value. Thus, whenever the time dynamically changes, it is not fetched from the backend. It happens on the frontend only, based on Date.now() - file.createdAt(). 


However, in case of downloads display, it has to be fetched from the backend (as far as my knowledge goes, I may be wrong though) and thus it needs to be refreshed everytime for displaying the same. 

Have a look at the preview for more details.

shikhar-scs:
@dannycoates mind reviewing?

dannycoates:
> mind reviewing?

yep. please push your changes to this PR's branch.

shikhar-scs:
I'll push them in a while, but is the content of the  gif file good

shikhar-scs:
@dannycoates I had already pushed my commits when the PR was in a closed state. Also, the changes do reflect here. Though it isn't showing a new commit. Please go to `file diff` to see the changes. 


Also the CI tests fail because of some similar reasons.

app/fileManager.js
```diff
@@ -68,6 +82,8 @@ export default function(state, emitter) {
       if (!ok) {
```
dannycoates: It's wasteful to make two requests to the server. Try to remove the `exists` call and use the result of `getDLCounts` to also determine if the file no longer exists.

app/fileManager.js
```diff
@@ -59,17 +69,61 @@ export default function(state, emitter) {
...
+            fileDetail.dtotal = file.dtotal;
+            fileDetail.dlimit = file.dlimit;
+            return true;
+          } else {
+            return false;
+          }
+        }
+      }
+    }
+  }
```
shikhar-scs: This entire functions controls the possibilities of `rerender` which include whether a dlimit or dcount change has occurred.

All existing files are stored in `fileIdArray`.

Details of all `dcount` and `dlimit` of a given file is stored in `dlimitArray`. And then if it has changed and then rerender is called. Pretty easy to understand though.

app/fileManager.js
```diff
@@ -59,17 +69,61 @@ export default function(state, emitter) {
...
     const files = state.storage.files;
     let rerender = false;
+
     for (const file of files) {
-      const ok = await exists(file.id);
-      if (!ok) {
+      const receivedFile = await getDLCounts(file).then(function(data) {
+        return data;
+      });
+      console.log(receivedFile);
```
shikhar-scs: This has been the main problem. The `console.log()` statement here prints an undefined statement because of which no further execution happens. 

There is something wrong with the promise returned here as the `.then` function receives no data here. Please help .

public/locales/en-US/send.ftl
```diff
@@ -41,6 +41,8 @@ deleteFileButton = Delete file
 sendAnotherFileLink = Send another file
 // Alternative text used on the download link/button (indicates an action).
 downloadAltText = Download
+downloads = Downloads
+timeLeft = Time
 downloadFileName = Download { $filename }
```
shikhar-scs: changes have been made here

app/templates/fileList.js
```diff
@@ -12,7 +12,10 @@ module.exports = function(state, emit) {
           <th id="copy-file-list" class="center-col">${state.translate(
             'copyFileList'
           )}</th>
-          <th id="expiry-file-list">${state.translate('expiryFileList')}</th>
+          <th id="expiry-time-file-list" >${state.translate('timeLeft')}</th>
+          <th id="expiry-downloads-file-list" >${state.translate(
+            'downloads'
+          )}</th>
```
shikhar-scs: removed the hard coded strings

app/fileManager.js
```diff
@@ -51,6 +52,15 @@ function exists(id) {
...
 }
+*/
+
+function getDLCounts(file) {
+  return new Promise((resolve, reject) => {
+    const url = `/api/metadata/${file.id}`;
+    const receiver = new FileReceiver(url, file);
+    resolve(receiver.getMetadata(file.nonce));
+  });
+}
```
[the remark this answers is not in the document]
dannycoates: Aside from other changes you might want to make I'd like to see the style use `async/await` instead of `Promise`. i.e.

```js
async function getDLCounts(file) {
  const url = `/api/metadata/${file.id}`;
  const receiver = new FileReceiver(url, file);
  try {
    const result = await receiver.getMetadata(file.nonce)
  }
  // ...
  return // something?
}
```

app/fileManager.js
dannycoates: I'd prefer not to have `dlimitArray` and `fileIdArray` at the scope where they are. It keeps them alive longer than necessary. I think there's a way you can compare the old/new values without them. Maybe while you iterating over the files in `checkFiles`?

app/fileManager.js
```diff
@@ -59,17 +69,62 @@ export default function(state, emitter) {
...
     let rerender = false;
+
     for (const file of files) {
-      const ok = await exists(file.id);
-      if (!ok) {
+      const receivedFile = await getDLCounts(file).then(function(data) {
+        return data;
+      });
+      console.log(receivedFile);
+      if (receivedFile.dlimit === receivedFile.dtotal) {
```
dannycoates: I don't think we'll ever get here. The server deletes the file when dtotal = dlimit so we're likely to get a 404 instead.

app/fileManager.js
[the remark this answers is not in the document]
shikhar-scs: This again is not working. When i Try to logthe results, they are blank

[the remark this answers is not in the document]
shikhar-scs: > try {
    const result = await receiver.getMetadata(file.nonce)
  }

Apparently, `getMetaData` doesn't return anything and so the `result` is always null. Are you talking about `fetchMetaData`?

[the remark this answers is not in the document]
dannycoates: I'm not saying to use that exact code, it's just an example. I want you to use async/await.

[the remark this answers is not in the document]
shikhar-scs: oh

app/fileManager.js
shikhar-scs: > if (receivedFile.dlimit === receivedFile.dtotal) {

Then what condition should be used here ?

app/fileManager.js
shikhar-scs: Yeah that's true, I've implemented it there only now.

app/fileManager.js
dannycoates: It depends on how you decide to implement `getDLCounts`. Hint: `FileReceiver.getMetadata` throws an `Error(404)` when the file isn't found.

shikhar-scs: Oh thanx! Please have a look now

app/fileManager.js
```diff
@@ -51,6 +52,19 @@ function exists(id) {
...
+
+async function getDLCounts(file) {
+  const url = `/api/metadata/${file.id}`;
+  const receiver = new FileReceiver(url, file);
+  try {
+    await receiver.getMetadata(file.nonce);
+    return receiver.file;
+  } catch (e) {
+    console.log(e);
+    return null;
```
shikhar-scs: This should actually be `return true` for the condition to work 
```
const receivedFile = await getDLCounts(file);
 if (receivedFile) {
```

dannycoates: Is this a question? ;)

If this case returned `true` then both possible return values would be "truthy". i.e. `if(result)` would *always* be true.

Take a step back and look at the function, `getDLCounts(file)`. If you didn't know what it did, what would you expect / want it to return? Maybe a number? Maybe a Map or Object with multiple numbers (named dtotal and dlimit)? Now what about if the `file` wasn't valid anymore? Should it return a special value or throw an error? Think about how you'd like to *consume* this function, then try to make it behave that way.

shikhar-scs: Yeah I've got this. Thanx. Please have a look now.

app/fileManager.js
```diff
@@ -60,16 +73,63 @@ export default function(state, emitter) {
...
     for (const file of files) {
-      const ok = await exists(file.id);
-      if (!ok) {
+      const receivedFile = await getDLCounts(file);
+      if (!receivedFile) {
         state.storage.remove(file.id);
         rerender = true;
       }
+
+      if (
```
dannycoates: this should probably be `else if` otherwise we'll run into trouble when receivedFile is false.

app/fileManager.js
```diff
@@ -60,16 +73,63 @@ export default function(state, emitter) {
...
+        file.dlimit !== receivedFile.dlimit ||
```
dannycoates: Might `file` and `receivedFile` both be references to the same object?

shikhar-scs: `file` has been picked from state.storage.files whereas receivedFile is the one we got from `FileReceiver.js`.

Logging them on the console brings this 

sae


This shows them as exact identical. Even after doing a few downloads. What else can be the strategy to check for the same?

dannycoates: So we know that `file` and `receivedFile` are the same object here. We also know that `getDLCounts` gets the latest counts. We want to compare the latest counts to the previous counts. When (as in line in this function) do we know the previous counts, and can we remember them so we can compare them here?

It might help to step through it in the debugger

shikhar-scs: Are you hinting at `state.storage.getFileById(file.id)` ?

dannycoates: > Are you hinting at state.storage.getFileById(file.id) ?

nope :)

think about this sample code:
```js
function foo(obj) {
  obj.a = obj.a + 1;
  return obj;
}

const x = { a: 1 };
console.log(x.a);

const y = foo(x);

console.log(y.a, x.a);

// we want to compare the _old_ value of x.a with y.a
if (x.a !== y.a) {
  console.log('this will never happen because x === y');
}
// when do we know the _old_ value of x.a? (when it was 1)
```

shikhar-scs: Apologies for the delay.

In this code, according to me, the old value of x.a is known at the 6th line of the above code OR basically, just at the function call. An equivalent of this in my code should right before when I send a file to send . So that means the previous commit of comparing file and receiverFile should work. Though we have seen that it doesnt work. Would you suggest something different??

dannycoates:
```js
  async function checkFiles() {
    const files = state.storage.files;

    let rerender = false;

    for (const file of files) {
      const oldLimit = file.dlimit;
      const oldTotal = file.dtotal;
      const receivedFile = await getDLCounts(file);
      if (!receivedFile) {
        state.storage.remove(file.id);
        rerender = true;
      }

      if (
        oldLimit !== receivedFile.dlimit ||
        oldTotal !== receivedFile.dtotal
      ) {
        rerender = true;
      }
    }

    if (rerender) {
      render();
    }
}
```

shikhar-scs: Thanks a lot @dannycoates. I just never realised that the contents of `file` too are changed after it is passed. I'll implement the same. are there any other changes needed?

shikhar-scs: @dannycoates I've made these changes. Are there any other changes to be made ?

app/fileManager.js
```diff
@@ -60,16 +73,63 @@ export default function(state, emitter) {
...
+            fileDetail.dtotal = file.dtotal;
+            fileDetail.dlimit = file.dlimit;
+            return true;
+          } else {
+            return false;
+          }
+        }
+      }
+    }
+  }*/
```
dannycoates: Please remove the commented out code

public/locales/en-US/send.ftl
```diff
@@ -41,6 +41,8 @@ deleteFileButton = Delete file
 sendAnotherFileLink = Send another file
 // Alternative text used on the download link/button (indicates an action).
 downloadAltText = Download
+downloads = Downloads
```
dannycoates: Please rename this to `downloadsFileList`

public/locales/en-US/send.ftl
```diff
@@ -41,6 +41,8 @@ deleteFileButton = Delete file
...
+timeLeft = Time
```
dannycoates: Please rename this to `timeFileList`

app/fileManager.js
shikhar-scs: yep sure I'll make all these changes :+1:

shikhar-scs:
@dannycoates I've made all the changes. Please review.

I'll squash them once they are ready for merge :smile:

dannycoates:
@shikhar-scs this looks good! Go ahead and squash it

shikhar-scs:
@dannycoates due to a poor rebase earlier, it led to a bad squashing result. This got out of control. So i've filed a new PR at #721. Please have a look there. :)

PR #683, opened 2017-12-21
API changes to accommodate 3rd party clients

### LINKED ISSUE #618: Include nonce & password status in /api/exists response
Context: for 3rd party clients / I'm updating my python send client to support passwords https://github.com/ehuggett/send-cli/issues/4

As far as know, to check if a password is required it is currently necessary to parse the html from `/download/$fileID/` for the value of the `data-requires-password` attribute belonging to the span with id `dl-file` (below)

```html
<span id="dl-file"
   data-nonce="XYuAjx5mI85Ac5LkkpHL0g=="
   data-requires-password="1">Download Encrypted File</span>
```

It would be very convenient to avoid parsing html if possible, and i believe this would be true in other languages should anyone else write a client (I'm not aware of any?), so I'm suggesting that
1. password requirements are indicated in the headers or json response body to make them more accessible (parsing json is already a requirement for any client)
2. if this is added to `api/exists`, also include the nonce or WWW-Authenticate header 
2.1 This would allow `api/exists` to return all of the required information for the client to then call `api/metadata` (no need to request `download/$fileID`)

I would be happy to submit a PR for this, but its not clear to me if it would be accepted

ehuggett:
Resolves #618

### DIFF
--- a/server/routes/exists.js
+++ b/server/routes/exists.js
@@ -14,7 +14,7 @@ module.exports = async (req, res) => {
     const meta = await storage.metadata(id);
     res.set('WWW-Authenticate', `send-v1 ${meta.nonce}`);
     res.send({
-      password: meta.pwd
+      password: meta.pwd !== '0'
     });
   } catch (e) {
     res.sendStatus(404);
--- a/server/routes/exists.js
+++ b/server/routes/exists.js
@@ -11,8 +11,11 @@ module.exports = async (req, res) => {
   }
 
   try {
-    await storage.exists(id);
-    res.sendStatus(200);
+    const meta = await storage.metadata(id);
+    res.set('WWW-Authenticate', `send-v1 ${meta.nonce}`);
+    res.send({
+      password: meta.pwd
+    });
   } catch (e) {
     res.sendStatus(404);
   }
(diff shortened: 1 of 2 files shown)


server/routes/exists.js
```diff
@@ -11,8 +11,11 @@ module.exports = async (req, res) => {
   }
 
   try {
-    await storage.exists(id);
-    res.sendStatus(200);
+    const meta = await storage.metadata(id);
+    res.set('WWW-Authenticate', `send-v1 ${meta.nonce}`);
+    res.send({
+      password: meta.pwd
```
dannycoates: `meta.pwd` is a string iirc, '0' or '1'. It might be better to make the public type a boolean (or really anything besides '0' and '1') 😆

server/routes/metadata.js
```diff
@@ -29,6 +29,8 @@ module.exports = async function(req, res) {
     const ttl = await storage.ttl(id);
     res.send({
       metadata: meta.metadata,
+      dtotal: meta.dl,
+      dlimit: meta.dlimit,
```
dannycoates: same as `meta.pwd` -- `dtotal` and `dlimit` should be numbers

server/routes/exists.js
ehuggett: oh dear that was a mess, string zero for false 😱 (I was ignorant about how redis stores values and assumed I would be retrieving an int, but I should have noticed the quotes in the output/json).

ehuggett:
Perhaps I should remove the changes to the metadata until #661 is resolved?

Issue #217, opened 2017-07-18
Delete button does not bring up a confirmation dialog


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 54.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Start the Firefox browser and navigate to "https://send.dev.mozaws.net/" page.
2. Upload a file and click the "Delete file" button.
3. Observe the behavior.

[Expected result]:
- A confirmation dialog appears asking the user to confirm the deletion. 

[Actual results]:
- The file is deleted.

[Notes]:
- For an action as important as deleting something, I think it would be best to have a confirmation dialog to delete a file. Especially on the "Copy link" page.
- Here is a screen recording with the issue:
![no confirmation dialog for delete](https://user-images.githubusercontent.com/18304511/28315923-379ea016-6bc9-11e7-81f6-da10bc65c40a.gif)

pdehaan:
**Proposal:** Make the button background color red, or something equally intimidating to possibly deter people from clicking it. Ref #270

gene1wood:
Really just any indication as to whether or not the delete was successful would be great here.

himanish-star:
@pdehaan , I'm on it

shikhar-scs:
I guess we still are missing a pop-up for delete button on the set password page. May I work on that ?

himanish-star:
@shikhar-scs , I have done that too !

SoftVision-CiprianMuresan:
I have to reopen this issue, while the confirmation dialog does appear when the button is clicked, the buttons (Cancel, Yes) do not work and the dialog can't be dismissed by clicking away from it:

![delete dialog buttons do not work](https://user-images.githubusercontent.com/18304511/34717061-72b883e2-f53a-11e7-9a5e-aa9170d46604.gif)

himanish-star:
@SoftVision-CiprianMuresan , you can see my PR all works fine.  It may be due to some small errors with new commits that it's not working. But don't worry I'll fix it in a while

SoftVision-CiprianMuresan:
@himanish-star, it seems to work fine on Chrome, Safari and Opera, but it does not work on Firefox Release, Beta or Nightly.

himanish-star:
@SoftVision-CiprianMuresan , OK I'll see to this

himanish-star:
@SoftVision-CiprianMuresan , are you now satisfied with the changes.

[...]

PR #684, opened 2017-12-23
delete btn popup attached

himanish-star:
fixes: #217 

As requested by @pdehaan and @SoftVision-CiprianMuresan , I have changed the background color of delete btn to red and attached a popup message . Although I believe the hovering effects are not that great. If you suggest me then I'll change it accordingly  

@dannycoates If you don't mind could you review my PR

**Below is a Preview Video**

ezgif com-video-to-gif 1

### DIFF
--- a/app/templates/share.js
+++ b/app/templates/share.js
@@ -65,7 +65,23 @@ module.exports = function(state, emit) {
       <button id="delete-file"
         class="btn"
         title="${state.translate('deleteFileButton')}"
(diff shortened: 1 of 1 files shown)


himanish-star:
@dannycoates , Sure I'll revert the colours back.

dannycoates:
I'd suggest changing `popupText` to something like:

```js
const popupText = document.querySelector('.popuptext')
```

himanish-star:
On it

dannycoates:
Thanks @himanish-star 😄

Issue #689, opened 2017-12-28
The string 'Firefox Send' is hardcoded everywhere


bmintz:
I'm trying to make my own fork of Firefox Send that's more privacy/anonymity oriented (ie no Google Analytics, no "Experiments" code, etc). I also wanted to change the title of it. I'd think that just changing state.title to `SendCrew` would be enough, but alas, not enough stuff uses it.

Even though I removed metrics, I still thought you'd like to know. Also this one would be a real pain to change cause it has no access to server/state.js or server/config.js.
https://github.com/mozilla/send/blob/46a3933dbba92a0fb2c42c0f6faabb72cc3f3fb4/app/metrics.js#L11-L15
https://github.com/mozilla/send/blob/46a3933dbba92a0fb2c42c0f6faabb72cc3f3fb4/app/fileManager.js#L85-L92
https://github.com/mozilla/send/blob/46a3933dbba92a0fb2c42c0f6faabb72cc3f3fb4/app/routes/index.js#L16-L18

wresuolc:
Thanks for the bug, but I'm going to close this -- that's the name of the product and since it's an experiment we're keeping the code simple and trying to move quickly on features.  Thanks!

PR #690, opened 2017-12-28
Use `state.title` in more places

bmintz:
See #689

### DIFF
--- a/app/fileManager.js
+++ b/app/fileManager.js
@@ -86,7 +86,7 @@ export default function(state, emitter) {
     document.addEventListener('blur', () => (updateTitle = true));
     document.addEventListener('focus', () => {
       updateTitle = false;
-      emitter.emit('DOMTitleChange', 'Firefox Send');
+      emitter.emit('DOMTitleChange', state.title);
     });
     checkFiles();
   });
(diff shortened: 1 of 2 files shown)


app/fileManager.js
```diff
@@ -86,7 +86,7 @@ export default function(state, emitter) {
     document.addEventListener('blur', () => (updateTitle = true));
     document.addEventListener('focus', () => {
       updateTitle = false;
-      emitter.emit('DOMTitleChange', 'Firefox Send');
+      emitter.emit('DOMTitleChange', state.title);
```
dannycoates: This isn't what we need here. The `DOMTitleChange` event sets `state.title` to the passed argument, so this effectively does nothing. A good way to do this is to add a config parameter to the server and pass it through to the frontend.

dannycoates:
Accomplishing #689 needs more work than this, and unfortunately this patch doesn't really help towards it. Factoring out the product name is not a priority for us right now, but we'd consider a more complete patch from the community.

app/fileManager.js
bmintz: Could you elaborate on that alternative?

dannycoates: We'll probably need:
- a new config parameter / variable in `server/config.js` for `PRODUCT_NAME`
- add said variable to `server/routes/jsconfig.js`
- add that variable to `state` in `app/main.js`
- also add it to `server/state.js`
- change the hard coded strings

Issue #687, opened 2017-12-25
Suggestion: An ability to change password


arunsathiya:
Currently when you upload a file, you have an option to set a password. Type a letter wrong, and you cannot change the password.

It would make sense to have an ability to change the password after setting one.

Especially when the file you upload is very large. You cannot upload the file again, just to select a new password.

ehuggett:
I think the **server** currently lets you do this (see send/server/routes/password.js, an authenticated request overwrites the auth key) but the **client** has no UI for it

Users may also expect to be able to unset/remove the password if they can edit it (which the server does NOT currently allow).

himanish-star:
@ehuggett  could I work on this issue

himanish-star:
@ehuggett , I am stuck I need some help. I have created the UI to reset password and I have used the same part of code as the add password button does. So the submission uses the same mechanism as add Password. 

After two days I still can't find my way through the authentication.
I always get a 401 error in this part as the verification fails. Could you help me ? I just want to know how to get the Authentication right. Below is the code snippet where the verification fails

         if (!verifyHash.equals(Buffer.from(auth, 'base64'))) {
         res.set('WWW-Authenticate', `send-v1 ${meta.nonce}`);
         return res.sendStatus(401);

ehuggett:
(as you can see) I just referenced this issue from my own project (a 3rd party client written in python),  I had previously said I "think" the server allows this, I now _know_ that the server allows it as I was able to change the password for a previously uploaded file with some modifications to my client

---

> @ehuggett could I work on this issue

I have no rights or control within this project (or Mozilla) so that's nothing to do with me!

> @ehuggett , I am stuck I need some help.

Sure, if you push what you have done so far to https://github.com/himanish-star/send I (and perhaps others) would be able to see/comment on it to try and help

> I don't know how to get the authHeader correct so that the verification passes

My first guess (as i can't see what you've done) would be your trying to authenticate with the old key instead of the new one?

1. [The initial authKey is generated from secretKey](https://github.com/mozilla/send/blob/797cfcb98de29104eedbd4b0811d5779a0155d10/app/fileSender.js#L198) and is sent to the server during the file upload. The client uses this key to authenticate if no password has been set.
2. When the user sets a password for the first time
2.1. the client generates [newAuthKey](https://github.com/mozilla/send/blob/797cfcb98de29104eedbd4b0811d5779a0155d10/app/fileSender.js#L296) using the password
[...]

PR #686, opened 2017-12-23
Hide password while Typing and after Entering: Fixes #670 

shikhar-scs:
As suggested by @dannycoates #673 all the requested changes have been made. Here's a preview
(its a gif file, please be patient while it loads)
Also Fixes #670 

ezgif com-video-to-gif

Please review

PR #679, opened 2017-12-21
changed font to sans sherif: Solves #676

shikhar-scs:
The font as marked out by @johngruen in #676 was different (Fira sans) and it has now been set to **Sans sherif** as required.

Issue #324, opened 2017-07-26
[idea] use Push API to inform senders when the download is complete


dannycoates:
This might actually be a *nice* use case for Push!

https://developer.mozilla.org/en-US/docs/Web/API/Push_API

AyoAlfonso:
Is this issue still open? Have you found a solution to this ?

pdehaan:
@AyoAlfonso I believe @abhinadduri has a work-in-progress PR for this at #379; _"WIP: added web push notifications for download completion"_.

ehuggett:
Has any thought been given to how this feature could be abused if the file receiver is not asked for permission to do so? I think it should be completely optional, like an email read receipt. From a quick read of the pull request it looks like it's triggered by the download of the file itself, not the page that displays the download link, so could the receiver's client signal their desire to opt-in with the request they use to retrieve it? (https://send.firefox.com/assets/download/de7ec7ab1e?notify_sender=true).

Does this rely upon the service knowing the name of the encrypted file? (n.b. I have objected to this in https://github.com/mozilla/send/issues/417) I'm wondering if it could be achieved with only the fileid being sent in the push notification from the server to the sender. The sender would have to keep a local record of previously uploaded fileid/filename(s) (i assume this does not happen already?) and use this data in the displayed notification. 

There would still be some potential problems with this, but they might be acceptable in this context as its unavoidable that the service can compromise the security of the transfer when it is being sent or received (as the users browser will execute whatever JavaScript is sent to it, stealing the encryption key etc) but it should at least be possible to ensure that this cannot happen after it has been completed.

Its probably unavoidable that the sender would store the metadata after the download has been completed. The push notification could perhaps trigger the deletion of it, but it may not be received. The best that would probably be achievable is if it could expire after 24 hours (as it can't be downloaded after that point anyway), limiting the risk to the 24 hour period after the upload. But if the expiry is not handled by the browser (like a cookie, i presume local Storage has an expiry and works the same?) then it could still potentially be available the next time the user uses the service (days, weeks etc later).

ehuggett:
Notifying the file sender that the file was not downloaded might be of interest to you, Its a lot less revealing and could be done either as an alternative or in addition to the on-download notification.

[...]

PR #694, opened 2018-01-03
Passwords can now be changed (#687)

himanish-star:
fixes: #687 

@ehuggett , Thanks for the help I have done the required changes

simplescreenrecorder-2018-01-03_17 14 13

### DIFF
--- a/app/fileSender.js
+++ b/app/fileSender.js
@@ -262,7 +262,7 @@ export default class FileSender extends Nanobus {
     return this.uploadFile(encrypted, metadata, new Uint8Array(rawAuth));
   }
 
-  static async setPassword(password, file) {
+  static async setPassword(existingPassword, password, file) {
     const encoder = new TextEncoder();
     const secretKey = await window.crypto.subtle.importKey(
       'raw',
@@ -293,6 +293,28 @@ export default class FileSender extends Nanobus {
       false,
       ['deriveKey']
     );
+    const oldPwdkey = await window.crypto.subtle.importKey(
+      'raw',
+      encoder.encode(existingPassword),
+      { name: 'PBKDF2' },
+      false,
+      ['deriveKey']
+    );
+    const oldAuthKey = await window.crypto.subtle.deriveKey(
+      {
+        name: 'PBKDF2',
+        salt: encoder.encode(file.url),
+        iterations: 100,
+        hash: 'SHA-256'
+      },
+      oldPwdkey,
+      {
+        name: 'HMAC',
+        hash: 'SHA-256'
+      },
+      true,
+      ['sign']
+    );
     const newAuthKey = await window.crypto.subtle.deriveKey(
       {
         name: 'PBKDF2',
@@ -310,10 +332,18 @@ export default class FileSender extends Nanobus {
     );
     const rawAuth = await window.crypto.subtle.exportKey('raw', newAuthKey);
     try {
-      await sendPassword(file, authKey, rawAuth);
+      if (existingPassword) {
+        await sendPassword(file, oldAuthKey, rawAuth);
+      } else {
+        await sendPassword(file, authKey, rawAuth);
+      }
     } catch (e) {
       if (e.message === '401' && file.nonce !== e.nonce) {
-        await sendPassword(file, authKey, rawAuth);
+        if (existingPassword) {
+          await sendPassword(file, oldAuthKey, rawAuth);
+        } else {
+          await sendPassword(file, authKey, rawAuth);
+        }
       } else {
         throw e;
       }
--- a/app/templates/share.js
+++ b/app/templates/share.js
@@ -10,10 +10,45 @@ function passwordComplete(state, password) {
   const el = html([
     `<div class="selectPassword">${state.translate('passwordResult', {
       password: '<pre></pre>'
-    })}</div>`
+    })}
+    <button id="resetButton">Reset</button>
+    <form id='reset-form' class="setPassword hidden" data-no-csrf>
+      <input id="unlock-reset-input"
+        class="unlock-input input-no-btn"
+        maxlength="64"
+        autocomplete="off"
+        placeholder="${state.translate('unlockInputPlaceholder')}">
+      <input type="submit"
+        id="unlock-reset-btn"
+        class="btn btn-hidden"
+        value="Reset Password"/>
+    </form>
+    </div>`
   ]);
-  el.lastElementChild.textContent = password;
+  el.firstElementChild.textContent = password;
+  el.querySelector('#resetButton').onclick = toggleResetInput;
+  el.querySelector('#unlock-reset-input').oninput = inputChanged;
   return el;
+}
+
+function inputChanged() {
+  const resetInput = document.getElementById('unlock-reset-input');
+  const resetBtn = document.getElementById('unlock-reset-btn');
+  if (resetInput.value.length > 0) {
+    resetBtn.classList.remove('btn-hidden');
+    resetInput.classList.remove('input-no-btn');
+  } else {
+    resetBtn.classList.add('btn-hidden');
+    resetInput.classList.add('input-no-btn');
+  }
+}
+
+function toggleResetInput(event) {
+  const form = event.target.parentElement.querySelector('form');
+  if (form.style.visibility === 'hidden' || form.style.visibility === '')
+    form.style.visibility = 'visible';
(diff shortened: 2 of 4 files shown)


himanish-star:
> If no password has been set then `existingPassword` is undefined, this error prevents the password from being set the first time.

Nope this doesn't happen I believe. @ehuggett , you can also see the gif and over there to the password was being set the first time and no error shows. Errors show up only when we try to reset the password. The reset procedure fails once with status 401 and then proceeds with status 200

ehuggett:
> Nope this doesn't happen I believe. @ehuggett , you can also see the gif(...)

I get the error when using Firefox Developer Edition 58.0b13 (64-bit) & Firefox ESR 52.5.2 (64-bit)
, Your PR does seem to work in chromium version 63.0.3239.84. 

note: the 'reset' button is not well formatted in Firefox either

> The reset procedure fails once with status 401 and then proceeds with status 200

I see that now. sendPassword was only updating the stored nonce if the request failed, my changes to have it always update the file nonce seem clean enough so I've opened a PR to have you include them in your branch https://github.com/himanish-star/send/pull/1 (which should cause it to be included here, so you should test it yourself before merging it).

himanish-star:
@ehuggett , thanks for the help. This issue has really helped me get a deeper understanding of the mozilla/send code base. So I need to correct the logic of existingPassword, got to format the reset Button and got to merge your branch onto mine. Ok I'm working on it. I'll make these changes later on , as I now got to go to college. If there are any other changes you want me to incorporate please do tell me.

shikhar-scs:
Also @ehuggett  would you mind reviewing these PRs #695 #686 #681 #679.
I have @dannycoates too but he seems to unavailable since a long time.

dannycoates:
@shikhar-scs @himanish-star I appreciate your enthusiasm, but please chill 😉 

I will be able to review your PRs soon. I'm the only current maintainer who can merge them.

himanish-star:
@dannycoates,  sorry for the inconvenience 😅

shikhar-scs:
@dannycoates, yeah Sorry for the inconvenience .

himanish-star:
@ehuggett , Now you can have a look at the changes . I have fixed all the bugs and have made the requested changes and I have also merged your branch into my branch

himanish-star:
@dannycoates , shall I resolve the conflicts ?

dannycoates:
@himanish-star yes, you can resolve conflicts, but it will probably be a little while longer before this feature gets reviewed / merged.

himanish-star:
@dannycoates , ok I will do this in a later part of the day

himanish-star:
@dannycoates , I have fixed the conflicts and things are back to normal.

app/fileSender.js
```diff
@@ -310,10 +330,18 @@ export default class FileSender extends Nanobus {
     );
     const rawAuth = await window.crypto.subtle.exportKey('raw', newAuthKey);
     try {
-      await sendPassword(file, authKey, rawAuth);
+      if (existingPassword) {
+        await sendPassword(file, oldAuthKey, rawAuth);
+      } else {
+        await sendPassword(file, authKey, rawAuth);
+      }
```
dannycoates: Just a nit on style... I'd prefer something like:
```js
const aKey = existingPassword ? oldAuthKey : authKey;
await sendPassword(file, aKey, rawAuth);
```

app/templates/share.js
```diff
@@ -10,20 +10,59 @@ function passwordComplete(state, password) {
...
+    resetInput.classList.remove('input-no-btn');
+  } else {
+    resetBtn.classList.add('btn-hidden');
+    resetInput.classList.add('input-no-btn');
+  }
+}
+
+function toggleResetInput(event) {
+  const form = event.target.parentElement.querySelector('form');
+  if (form.style.visibility === 'hidden' || form.style.visibility === '')
```
dannycoates: Another style nit: no `if`s without brackets

app/fileSender.js
```diff
@@ -18,13 +18,11 @@ async function sendPassword(file, authKey, rawAuth) {
     const xhr = new XMLHttpRequest();
     xhr.onreadystatechange = () => {
       if (xhr.readyState === XMLHttpRequest.DONE) {
+        const nonce = xhr.getResponseHeader('WWW-Authenticate').split(' ')[1];
```
dannycoates: There are cases where this header is not set (status == 404 for example), so would throw an exception here

app/templates/share.js
```diff
@@ -99,6 +138,21 @@ module.exports = function(state, emit) {
   </div>
   `;
 
+  if (div.querySelector('#reset-form'))
+    div.querySelector('#reset-form').onsubmit = resetPassword;
+
+  function resetPassword(event) {
+    event.preventDefault();
+    const existingPassword = document.querySelector('.passwordOriginal')
+      .innerText;
```
himanish-star: This is regarding indentation . Shall I do this 
 ```javascript
const existingPassword = document.querySelector('.passwordOriginal')
        .innerText;
```
or this
```javascript
 const existingPassword = document.querySelector('.passwordOriginal').innerText;
```

app/fileSender.js
```diff
@@ -18,13 +18,11 @@ async function sendPassword(file, authKey, rawAuth) {
...
+        file.nonce = nonce;
         if (xhr.status === 200) {
           return resolve(xhr.response);
         }
-        if (xhr.status === 401) {
```
himanish-star: Do you want me to revert this removed code back. If in case the `status=401` then the below code would help to reset the `file.nonce`
```javascript
if (xhr.status === 401) {
 const nonce = xhr.getResponseHeader('WWW-Authenticate').split(' ')[1];
 file.nonce = nonce;
}
```

app/templates/share.js
dannycoates: In general try to keep it on one line, but our auto-formatter `prettier` might change it.

app/templates/share.js
```diff
@@ -10,20 +10,61 @@ function passwordComplete(state, password) {
   const el = html([
     `<div class="selectPassword">${state.translate('passwordResult', {
       password: '<pre></pre>'
-    })}</div>`
+    })}
+    <button id="resetButton">Reset</button>
```
dannycoates: I missed this earlier... The "Reset" text here needs to be localized. We need to add a string to `public/locales/en-US/send.ftl` and use `state.translate` like the other strings here. I also think "Change" is a better word than "Reset" in this case.

app/templates/share.js
```diff
@@ -10,20 +10,61 @@ function passwordComplete(state, password) {
...
+    <form id='reset-form' class="setPassword hidden" data-no-csrf>
+      <input id="unlock-reset-input"
+        class="unlock-input input-no-btn"
+        maxlength="64"
+        autocomplete="off"
+        placeholder="${state.translate('unlockInputPlaceholder')}">
+      <input type="submit"
+        id="unlock-reset-btn"
+        class="btn btn-hidden"
```
dannycoates: Does it work if you set `onclick=${toggleResetInput}` here instead of with the querySelector below? If so it's the preferred style. (also same with others)

app/templates/share.js
himanish-star: Yes , I know that it has to be localised. But I thought I would have to do it in a number of different languages . Thus I had left it alone. I'll make the changes soon.

app/templates/share.js
himanish-star: Yes, I know that this is the preferred style, but I had tried to do it this way but it hadn't worked. I'll try again and if it doesn't work , I'll ask you for help :smiley:

himanish-star: Nope , using `onclick=${toggleResetInput}` doesn't work :disappointed:. Can you try the same thing out once @dannycoates

himanish-star: > We need to use the function form here so that the password text gets escaped properly. It's not a big deal.

So you mean we can still write `onclick=${toggleResetInput}`. Ok, I'll try searching more over the net of how to do this. Meanwhile if you can suggest me of how this is done, then I will surely rectify my mistake

himanish-star:
@dannycoates , Thank for merging. Can I take up issue #324 ? :smiley:

app/templates/share.js
dannycoates: No, no, I was just commenting on why we needed to use html() instead of html``

Your code is fine :)

PR #695, opened 2018-01-03
Show Warning, Cancel and Redirect on size > 2GB ; fixes #578

### LINKED ISSUE #578: The "File size" error message is no longer displayed when uploading a file over 2GB
[Affected versions]:
- Firefox 55.0.3 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Steps to reproduce]:
1. Open the browser and navigate to "send.stage.mozaws.net" page.
2. Upload a large file over 2GB.
3. Observe the browser behavior.

[Expected result]:
- An error message is displayed to warn you that the file size limit is 2GB.

[Actual results]:
- You are redirected to the "Something went wrong" page.

[Notes]:
- I recall that there was an error message that was displayed when you tried to upload a large file (over 2GB), but it was probably lost somewhere.
- Here is a screen recording with the issue:
over 2b

shikhar-scs:
As mentioned by @SoftVision-CosminMuntean and @johngruen at #578, this PR resolves issues related to file uploads of sizes > 2GB. It displays the warning message, cancels the upload and further redirects back to the home page. 

The following preview has been generated for file sizes > 200MB as my laptop crashes for any file having size > 1 GB. However, this limit can be very easily altered as per our needs.

Here is the preview (its a gif file so please be patient while it loads)

filesize

@dannycoates @SoftVision-CosminMuntean @johngruen Please review and suggest changes.

### DIFF
--- a/app/templates/welcome.js
+++ b/app/templates/welcome.js
@@ -1,7 +1,8 @@
+/* global MAXFILESIZE */
 const html = require('choo/html');
 const assets = require('../../common/assets');
 const fileList = require('./fileList');
-const { fadeOut } = require('../utils');
+const { bytes, fadeOut } = require('../utils');
 
 module.exports = function(state, emit) {
   const div = html`
@@ -64,6 +65,11 @@ module.exports = function(state, emit) {
     if (file.size === 0) {
       return;
     }
+    if (file.size > MAXFILESIZE) {
+      window.alert(state.translate('fileTooBig', { size: bytes(MAXFILESIZE) }));
+      return;
+    }
+
     await fadeOut('page-one');
     emit('upload', { file, type: 'click' });
   }
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 dist
+.idea


shikhar-scs:
> A better location to check the file size is in welcome.js. We can simply show an alert return early from that function.

Yeah, even I realized that right now. I'll have a look at it and make the necessary changes here itself.

shikhar-scs:
@dannycoates Although `welcome.js` is a much better location to check for the file size, so is the cancellation procedure. As it is done in case of `file.size===0`, all it takes is just a return statement to be executed. 

However displaying the relevant warning to the user in such a case is what is causing the problem. Though I've managed to do that using some script statements in the `welcome.js` file, its not looking good. I can do that if you want to OR maybe let it be as it is, as displayed in the `gif` file above as it equally solves the problem too. Whatsay?

dannycoates:
> However displaying the relevant warning to the user in such a case is what is causing the problem.

Let's have the check in `welcome.js`. A javascript `alert(msg)` is enough ui for this. There's already a localization string for this named `fileTooBig`.

shikhar-scs:
@dannycoates Ok, I'll put that in place. Should I also include one for a 0 byte file ?

dannycoates:
> Should I also include one for a 0 byte file ?

That one can stay how it is

shikhar-scs:
OK no problem :+1:

shikhar-scs:
@dannycoates I've now finally introduced alert warnings. I guess this might be it. Here's a preview 

The demo has been performed on a file size of 200MB (though the final commit contains the limit for 2GB) , reasons for which have been stated above. 


upload2gb

Please review

app/templates/upload.js
```diff
@@ -4,7 +4,26 @@ const { bytes } = require('../utils');
...
+      )}" onclick=${cancel}>${state.translate('uploadingPageCancel')}</button>
+       <script>
+       setTimeout(function(){ document.getElementById('cancel-upload').click(); }, 2000);
+     </script>
+    </div>
+  </div>
+  `;
+    return fileSizeExceededDiv;
+  }
+*/
```
dannycoates: please delete the commented code

app/templates/welcome.js
```diff
@@ -64,6 +64,14 @@ module.exports = function(state, emit) {
     if (file.size === 0) {
       return;
     }
+    if (file.size > 2000000000) {
```
dannycoates: please use the `MAXFILESIZE` global variable instead of `2000000000`

app/templates/welcome.js
```diff
@@ -64,6 +64,14 @@ module.exports = function(state, emit) {
...
+      window.alert(
```
dannycoates: please use `state.translate('fileTooBig', { size: bytes(MAXFILESIZE) })` instead of the hardcoded string. Where `bytes` is from `app/utils.js`

shikhar-scs:
I've made the required changes @dannycoates. Please review.

app/templates/welcome.js
```diff
@@ -1,5 +1,7 @@
 const html = require('choo/html');
 const assets = require('../../common/assets');
+const bytes = require('../utils');
+const config = require('../../server/config');
 const fileList = require('./fileList');
 const { fadeOut } = require('../utils');
```
dannycoates: add `bytes` here `const { bytes, fadeOut } = require('../utils');` instead of a separate `require`

app/templates/welcome.js
```diff
@@ -59,11 +61,17 @@ module.exports = function(state, emit) {
 
   async function upload(event) {
     event.preventDefault();
+    const MAXFILESIZE = config.max_file_size;
```
dannycoates: not quite. The `server` path in the require is a clue that we shouldn't use that file here. `MAXFILESIZE` gets set globally in the browser window by https://send.firefox.com/jsconfig.js so we can use it directly here.

shikhar-scs: Using the variable directly gives the following error 
```
67:21  error    'MAXFILESIZE' is not defined  no-undef
  68:7   warning  Unexpected alert              no-alert
  68:64  error    'MAXFILESIZE' is not defined  no-undef
```

And this prevents any commits.
So, we need to do something different I guess

dannycoates: This is the linter complaining. Try adding `/* global MAXFILESIZE */` as the first line of the file.

shikhar-scs: Correct, Its working now. I've pushed the commit please have a look.

shikhar-scs: Also, I would squash my commits once the PR is ready to be merged.

shikhar-scs:
All the commits have been squashed, this PR is ready to be merged.

dannycoates:
Thanks @shikhar-scs 😄

PR #701, opened 2018-01-11
improved popup for mobile display; Fixes #699

### LINKED ISSUE #699: The Delete dialog from the Link copy page is not correctly displayed in mobile view
[Affected versions]:
- Firefox 57 and up

[Affected Platforms]:
- All Platforms

[Steps to reproduce]:
1. Open the browser and resize it's width to the minimum.
2. Navigate to https://send.dev.mozaws.net and upload a file.
3. Click the delete button and observe the dialog.

[Expected result]:
- The dialog is fully displayed.

[Actual results]:
- The left side of the dialog box is cut off.

[Notes]:
- Here is a screen shot of the issue:
image

shikhar-scs:
As suggested by @SoftVision-CiprianMuresan at #699, the popup for mobile view has been centered and thus now falls in the visible window. Here's a preview

popup

Please review

### DIFF
--- a/assets/main.css
+++ b/assets/main.css
@@ -1181,4 +1181,12 @@ tbody {
     font-size: 13px;
     padding: 17px 5px 0;
   }
+
+  .popup .popuptext::after {
+    left: 125px;
+  }
+
+  #deletePopup {
+    left: 83px;
+  }
 }


shikhar-scs:
@dannycoates could you review please ?☺️

assets/main.css
```diff
@@ -1181,4 +1181,27 @@ tbody {
...
+    display: block;
+    width: 20px;
+    height: 20px;
+    transform: rotate(45deg);
+    border-radius: 0 0 5px;
+    border-right: 1px solid #d7d7db;
+    border-bottom: 1px solid #d7d7db;
+    border-left: 1px solid #fff;
+    border-top: 1px solid #fff;
+  }
```
dannycoates: It shouldn't be necessary to redefine all these css attributes here. We should only set the ones that need to be overridden or added.

assets/main.css
```diff
@@ -1181,4 +1181,27 @@ tbody {
...
+
+  #deletePopup {
```
dannycoates: Is this supposed to be only in this media block?

assets/main.css
shikhar-scs: Yeah that's true, I'll make the required changes

assets/main.css
shikhar-scs: Well, Yes. Changing its `left ` properties help us to shift it to the center. However, I'll remove the extra statements

shikhar-scs:
@dannycoates I've remoced the extra statements, and squashed the commits. Please review.

Issue #663, opened 2017-12-08
Restrict the banner from showing on unsupported browsers


SoftVision-CiprianMuresan:
[Notes]:
- Since the Download Firefox page is already displayed instead, we could do without the banner showing.

[Affected versions]:
- Firefox 57.0 and up

[Affected Platforms]:
- All Windows

[Steps to reproduce]:
1. Start the Edge browser with and navigate to "https://send.dev.mozaws.net/" page.
2. Observe the top part of the page.

[Expected result]:
- The download banner is not displayed.

[Actual results]:
- The download banner is displayed.

[Additional Notes]:
- Here is a screen recording with the issue:
![img_08122017_150638_0](https://user-images.githubusercontent.com/18304511/33767904-903e6a5c-dc2c-11e7-943e-ff781e04c8df.png)

BrunoScheufler:
I tried to reproduce it with Edge multiple times but the header didn't show up for me 😕 Does it still show up for you? I'm really curious about this know.

Softvision-CristinaBadescu:
Hello @BrunoScheufler ,
I've managed to reproduce this issue using Microsoft Edge 40.15063.674.0 on Windows 10x64, using English (United States) Language and US keyboard layout. Can you please provide the version of the Microsoft Edge you are using and also the set language and keyboard? Maybe due to this setting you didn't encounter this issue.
Below you can see a screenshot with the issue:
![send microsoft edge](https://user-images.githubusercontent.com/31244235/33876044-f66554d2-df2c-11e7-8ee5-14be66f34e23.png)

BrunoScheufler:
Hello @Softvision-CristinaBadescu ,
I've tried again using Win10 (1709) using the German language and German/DE keyboard layout. The header did not show up for me in any of the tests, that might be because of the language or some other factor. It looks like this to me:
![edge](https://user-images.githubusercontent.com/4772980/33892730-d3720e02-df59-11e7-8026-b7d6042fc70e.png)

dannycoates:
The banner is en-US only for now

BrunoScheufler:
@dannycoates well that explains it all, thank you 🙂

johngruen:
stretch!

himanish-star:
I would like to work on this issue . May I ?

dannycoates:
@himanish-star yes. The way I recommend approaching it is to check `state.route` in `app/routes/index.js` and not render the `fxPromo` if it equals `'/unsupported/:reason'`

[...]

PR #702, opened 2018-01-12
Restricted the banner from showing on unsupported browsers

himanish-star:
fixes : #663 
Here is a GIF of the changes
ezgif com-video-to-gif 11

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -7,10 +7,14 @@ const fxPromo = require('../templates/fxPromo');
 
 const app = choo();
 
+function showBanner(state) {
+  return state.promo && !state.route.startsWith('/unsupported/');
+}
+
 function body(template) {
   return function(state, emit) {
     const b = html`<body>
-      ${state.promo ? fxPromo(state, emit) : ''}
+      ${showBanner(state) ? fxPromo(state, emit) : ''}
       ${header(state)}
       <div class="all">
         <noscript>


app/routes/index.js
```diff
@@ -9,6 +9,9 @@ const app = choo();
 
 function body(template) {
   return function(state, emit) {
+    if (state.route.startsWith('/unsupported/')) {
+      state.promo = false;
```
dannycoates: This will toggle the banner off for the remainder of the session, which for almost all cases is still fine because folks who end up here aren't likely to go to any other routes, but we can be more precise, so we should be.

I suggest creating a helper function called `showBanner` (or something) that combines your comparison here with stuff on line 16 below

himanish-star: Ok, sure I'll do this

himanish-star:
> This will toggle the banner off for the remainder of the session

You mean that `state.promo` would get set as `false` throughout the session ? And that would cause the problem. 

I have made the changes but I'm afraid that I didn't quite achieve the changes you were expecting from me. I think I couldn't get the functionality right.

himanish-star:
> This will toggle the banner off for the remainder of the session

After going to route `/unsupported/:reason` the banner is hidden but then visiting to other routes should also hide them according to you but I can't see anything like this. 
ezgif com-video-to-gif 12

dannycoates:
> After going to route /unsupported/:reason the banner is hidden but then visiting to other routes should also hide them according to you but I can't see anything like this.

I didn't explain this very well. The only page that should hide the banner is the unsupported page, all others *should* show it. The way you have it now is correct.

app/routes/index.js
```diff
@@ -7,10 +7,14 @@ const fxPromo = require('../templates/fxPromo');
 
 const app = choo();
 
+function showBanner(route) {
+  return !route.startsWith('/unsupported/');
```
dannycoates: we still want to check `state.promo` so `state.promo &&`

himanish-star: > we still want to check state.promo.

Sorry, I missed upon that in a hurry

himanish-star: I have made the changes

app/routes/index.js
```diff
@@ -7,10 +7,14 @@ const fxPromo = require('../templates/fxPromo');
 
 const app = choo();
 
+function showBanner(state) {
+  return !(state.promo && state.route.startsWith('/unsupported/'));
```
dannycoates: The parameters look good, but the boolean logic isn't quite what we want. For example this return `true` when both are false.

The truth table we want is:

show | promo | unsupported
------|-------|-------------
true | true | false
false | true | true
false | false | true
false | false | false

himanish-star: I have made the changes.

himanish-star:
>  The only page that should hide the banner is the unsupported page, all others should show it.

I also meant the same. What I was trying to convey was that when I had pushed a commit for the first time where I had used the logic `state.promo = false` , you said this 
> This will toggle the banner off for the remainder of the session

But what I want to say is that (I hade also made a GIF above) \besides using `state.promo = false` (which I know is logically incorrect ) the banner was behaving the way it was meant to behave i.e. hidden in case of `unsupported` route and visible in case of other routes. I guess the reason behind this is that the state.promo is always reset to either 'grey' or 'blue'. So it doesn't make any difference if state.promo was set to false or not. 

Also as I was going through some issues I saw how `send` was working in `edge`. But whenever I open `send` in `edge` no matter which ever route I go to I always land up on `/unsupported/edge`

shikhar-scs:
> Also as I was going through some issues I saw how send was working in edge. But whenever I open send in edge no matter which ever route I go to I always land up on /unsupported/edge

Might be because you are working in chrome.

himanish-star:
@shikhar-scs , read clearly what I have commented

shikhar-scs:
Yep my bad.

dannycoates:
> banner was behaving the way it was meant to behave i.e. hidden in case of unsupported route and visible in case of other routes

It was in some cases but not others. If you were using the address bar to go to new urls, promo would get reset. If you were navigating from links on the page or the back button, promo would not get reset. Your first PR did not work as it should in the second case.

> So it doesn't make any difference if state.promo was set to false or not.

It does make a difference. In general its a bad idea to change state when unnecessary. There may be other parts of the system that are affected by it.

> Was my first commit reviewed for changes because of the fact that state.promo = false would set state.promo as false for the entire session ?

Yes

himanish-star:
> If you were navigating from links on the page or the back button, promo would not get reset.

@dannycoates , I am highly grateful to you for explaining me this. Thank you very much.

> Yes
Thanks for this too.
No more doubts

Issue #733, opened 2018-01-25
Download encrypted file displayed on downloading the file, and changes to Download "Filename" within seconds.


Radhraj:
Steps:
Go to https://send.firefox.com
Click on Select File to Upload
Upload a file.
Copy the URL
Hit the URL(Another browser)

Version:
Firefox 58.0.0

Expected Result:
The UI must display title as "Download Filename"

Actual Result:
The UI displays Download Encrypted File.
The title displayed above also changes back to Download File within few seconds.

![downloadencryptedfile_ondownload](https://user-images.githubusercontent.com/30893077/35381689-1846a492-01e3-11e8-9297-2f1db117697c.png)

himanish-star:
@dannycoates , Can I fix this issue?

dannycoates:
This is a side effect of the filename being encrypted. It gets fetched separately from the page and is then decrypted. While that's happening this generic message is shown.

There's a couple things I think we can do to make this better:

- change the message from "Dowload Encrypted File" to something else. @sevaan thoughts?
- don't show the download button until the filename is loaded

dannycoates:
@himanish-star you can work on the download button part if you like

himanish-star:
Thanks, I'll definitely do so

wresuolc:
Sevaan is going to provide UX support here

sevaan:
The "Download Encrypted File" is only up there for a split-second...plus, I noticed an additional issue, where the flying document image disappears and the word "Download" is flashed on the screen ([video](https://cl.ly/0u0X2E2a412Q)):

<img width="596" alt="image" src="https://user-images.githubusercontent.com/473531/35647256-f0e516d8-069f-11e8-8d3d-fd0a141541d1.png">

It's a lot of distracting movement. I think all of this could best be sorted by not showing any page content except an animated spinner/loader until the file name is fetched, then display everything at once. The footer and header and page background should still be shown, this is a quick loader for the content.

[Maybe one of these](http://samherbert.net/svg-loaders/)...I kind of like the second from the left on the bottom row (tailspin.svg). Or the far right in the middle (puff.svg). I think #000000 and centered in the page, right above where the flying envelope graphic is.

<img width="941" alt="image" src="https://user-images.githubusercontent.com/473531/35648466-d825c8f0-06a3-11e8-9118-f5b69d6af537.png">

sevaan:
Current implementation works fine for when there is a password.

dannycoates:
[...]

PR #736, opened 2018-01-27
don't show the download button until the filename is loaded

himanish-star:
fixes: #733 
@dannycoates, I have implemented the part you have requested for. Please review :smiley:

### DIFF
--- a/app/templates/preview.js
+++ b/app/templates/preview.js
@@ -38,6 +38,10 @@ module.exports = function(state, emit) {
       </button>
     </div>
   </div>`;
+  if (!state.transfer) {
+    const ele = action.querySelector('#download-btn');
+    ele.parentNode.removeChild(ele);
+  }
   if (fileInfo.pwd && !fileInfo.password) {
     action = downloadPassword(state, emit);
   } else if (!state.transfer) {


app/templates/preview.js
```diff
@@ -38,6 +38,10 @@ module.exports = function(state, emit) {
       </button>
     </div>
   </div>`;
+  if (!state.transfer) {
+    const ele = action.querySelector('#download-btn');
```
dannycoates: This breaks on the [server side](https://github.com/choojs/choo#server-rendering). The server also runs this code [from here](https://github.com/mozilla/send/blob/master/server/routes/pages.js#L30-L45), but there `action` is not a `DOMElement` it's just a string. So we end up in the `catch` case which falls through to the `notfound` code.

A good way to make it work on both the server and browser is to conditionally add the button in the `action` template string instead of removing it after.

himanish-star: Sure I will look into this in a while

dannycoates:
Thanks @himanish-star, I got a chance to fix this myself yesterday. I've integrated a fix for this in #737

himanish-star:
:disappointed:, sorry for the delay.Fine I'll fix another issue

himanish-star:
@dannycoates, could you assign me some other issue which you think I can solve? In the meanwhile even I am looking out for some issues which I can solve.

PR #739, opened 2018-01-30
added /api/info/:id route

### LINKED ISSUE #738: Split sender and receiver metadata
It might be better to have a separate server endpoint to get the metadata the sender needs instead of a single metadata endpoint that both sender and receiver can access.

dannycoates:
This removes the `dtotal` and `dlimit` from `/api/metadata` since the downloader doesn't need (and probably shouldn't have) them. It also allows us to remove the `authKey` from the sender's local storage.

fixes #738


ehuggett:
> since the downloader doesn't need (and probably shouldn't have)

does this mean there will be no way to warn a user (downloader) that this is the last download before the file is deleted? a boolean for 2 or more remaining downloads would probably be sufficient for the downloader

dannycoates:
> does this mean there will be no way to warn a user (downloader) that this is the last download before the file is deleted?

As is, this would do that. A boolean `finalDownload` or something seems ok though.

dannycoates:
I added `finalDownload` to the `/api/metadata` response

PR #759, opened 2018-02-15
Switch en-US FTL file to new syntax

flodolo:
The only change is for the comments sigil. 

CC @pike


Pike:
Should we add a term for Firefox Send and Test Pilot?

flodolo:
> Should we add a term for Firefox Send and Test Pilot?

Possibly, but not in this PR. I would like to keep the scope limited, i.e. no new/changed strings.

PR #760, opened 2018-02-15
refactored css: phase 1

dannycoates:
This mainly extracts the css into separate files and tries to make the naming scheme follow BEM.

There's still the opportunity to dedupe more (like a `.page` block) and extract colors for example. At least this passes `stylelint` again.


app/base.css
```diff
@@ -0,0 +1,217 @@
...
+
+.inputBtn {
+  background: #0297f8;
+  border-radius: 0 6px 6px 0;
+  border: 1px solid #0297f8;
+  color: white;
+  cursor: pointer;
+
+  /* Force flat button look */
+  appearance: none;
```
pdehaan: FWIW, this is the warning output from Stylelint:

```
app/base.css
 106:3  ⚠  Unexpected browser feature    plugin/no-unsupported-browser-features
           "css-appearance" is only                                            
           partially supported by                                              
           Firefox 52,56,57, Chrome                                            
           61,62, Safari 10,11,9.1,10.1
```

public/locales/en-US/send.ftl
```diff
@@ -113,3 +113,4 @@ enableJavascript = Please enable JavaScript and try again.
 expiresHoursMinutes = { $hours }h { $minutes }m
 # A short representation of a countdown timer containing the number of minutes remaining as digits, example "56m"
 expiresMinutes = { $minutes }m
+passwordIsSet = Password set
```
flodolo: Can you add some context in a comment about where this string shows up?

dannycoates:
@leroyjenkins r+

PR #781, opened 2018-03-06
Don't translate URL-safe chars, b64 is doing it for us

timvisee:
Send should not translate URL-safe characters to their non-safe variant when decoding a base64 string, as [`base64-js`](https://github.com/beatgammit/base64-js) is already doing that for us. The library seems to be doing it in a more performant manner, therefore I removed the manual translation from Send in this PR.

See the following snippet in the `base64-js` library:
https://github.com/beatgammit/base64-js/blob/master/index.js#L17-L20

### DIFF
--- a/app/utils.js
+++ b/app/utils.js
@@ -9,10 +9,9 @@ function arrayToB64(array) {
 }
 
 function b64ToArray(str) {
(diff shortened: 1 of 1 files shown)


dannycoates:
Cool! Thank you @timvisee 😄

PR #791, opened 2018-03-14
Replaced the selectbox with native HTML <select>

RCMainak:


app/templates/selectbox/index.js
```diff
@@ -6,54 +6,28 @@ module.exports = function(selected, options, translate, changed) {
...
-      ul.classList.add('selectbox__options--active');
-      const body = document.querySelector('body');
-      body.addEventListener('click', close);
-    }
-  }
+    </select>
+    <svg id="arrow" width="32" height="32">
+      <polygon points="8 18 17 28 26 18" fill="#0094fb"/>
+    </svg>
+  </span>`;
```
dannycoates: I'd suggest using `onchange` for the event handler, and `selected` on the selected option. Something like:

```js
html`
  <span class="select">
    <select id="${id}" onchange=${choose}>
      ${options.map(
        i =>
          html`<option value="${i}" ${i === selected ? 'selected' : ''}>${translate(i)}</option>`
      )}
    </select>
  </span>`;
```

dannycoates:
Great progress so far! Probably just more tweaking of the css to go. I'll be able to help out in another few days if you get stuck.

app/templates/selectbox/index.js
RCMainak: Of course, that is much more clearer.

dannycoates:
@RCMainak I see you've made changes. Are you done and ready for review?

RCMainak:
Yes, I made some changes. It's working properly but not sure about CSS. Please make recommendations, I'll make changes if needed.

dannycoates:
Thanks @RCMainak! I landed your changes after a couple tweaks on master.

PR #800, opened 2018-04-05
Initial user integration tests.

jrbenny35:
This adds some user integration tests to aide the SoftVision team a bit. Right now I have 3 tests.

```test_upload```: This will create a file and make sure it uploads by verifying a file uploads and is assigned a URL.

```test_download```: This will create a file, upload it and then download it making sure it is the same filename that was uploaded. We can expand this later to maybe check the sizes and such.

```test_progress```: This will create a file and make sure the progress bar shows up after it begins uploading.

These are python tests and use [Pipenv](https://docs.pipenv.org/) to manage dependencies as well as [tox](https://tox.readthedocs.io/en/latest/) as the virtualenv manager, and finally [pytest](https://docs.pytest.org/en/latest/) as the test runner. 

The tests build a docker container with the app and redis. I have a docker-compose override to expose the development port. 

I also added a URL for accessing the server besides ```localhost```, it is ```send.firefox.local```. This could be changed but it is needed to access the server within the selenium-firefox docker container.

I'll add some docs explaining more in the next PR, just wanted to get this up.

PR #806, opened 2018-04-17
trying wdio for selenium tests

dannycoates:


test/testServer.js
```diff
@@ -0,0 +1,23 @@
...
+      devRoutes(app, { middleware: wpm });
+      wpm.waitUntilValid(() => {
+        server = app.listen(8081, resolve);
+      });
+    });
+  },
+  onComplete: function() {
+    server.close();
+  }
+};
```
jrbenny35: This creates a web server that has all of the assets and images loading? That was one problem I ran into. I wonder if I could use this for my tests as I had to add a script to ignore the ```.dockerignore``` so I could run the  app with all the assets.

pdehaan:
@dannycoates Hey Danny, why did we close this? Was wdio/webdriver not working for us? Or did we come up with a different/better solution?

dannycoates:
@pdehaan closed in lieu of #800

PR #813, opened 2018-05-03
moved integration text execution to docker

dannycoates:
let's see if we can get testing on localhost to be the same as circleci


jrbenny35:
This is awesome! I have actually learned a good bit, and I am totally down for this setup.

docker-compose.yml
```diff
@@ -12,14 +12,15 @@ services:
...
-      - .:/send
-    working_dir: /send
-    expose:
-      - "4444"
     ports:
       - "5900"
       - "4444:4444"
     shm_size: 2g
+  integration-tests:
+    build: ./test/integration
```
jrbenny35: So you can build another image even though you pull an image? That's pretty neat.

test/integration/tox.ini
```diff
@@ -3,13 +3,11 @@ envlist = integration-tests, flake8
...
 
 [testenv]
-recreate=True
 skip_install = True
-passenv = DISPLAY MOZ_HEADLESS
 deps = -rpipenv.txt
 commands =
     pipenv install --skip-lock
-    pipenv run pytest -v --verify-base-url -n 2 --driver Firefox --html=send-test.html --self-contained-html {posargs}
+    pipenv run pytest -v --verify-base-url --driver Remote --capability browserName firefox --host selenium --html=send-test.html --self-contained-html {posargs}
```
jrbenny35: The driver is remote but isn't firefox within the same image?

dannycoates: Selenium is running in it's own (separate) container so that's why this is remote. You could instead base the integration-tests image on the selenium one but that'd make the new Dockerfile more complicated because it'd have to run multiple processes. I figured this way was a bit simpler.

test/integration/requirements.txt
```diff
@@ -0,0 +1,8 @@
+selenium==3.11.0
```
jrbenny35: Why the revert back to requirements.txt?

docker-compose.yml
```diff
@@ -12,14 +12,18 @@ services:
...
+  selenium:
     image: b4handjr/selenium-firefox
-    volumes:
-      - .:/send
-    working_dir: /send
-    expose:
-      - "4444"
     ports:
-      - "5900"
+      - "5900:5900"
```
jrbenny35: I'd rather keep this to just whatever open port can be found. I feel like using a specific port could cause problems.

test/integration/requirements.txt
dannycoates: I found runs to be about 15 seconds faster without pipenv. Unless there's a huge benefit to using it I'd prefer faster. I'm not very familiar with python's package management.

docker-compose.yml
dannycoates: we could make it an env variable, then anyone who does have problems on the default port can change it. Dynamic is less convenient in the common case.

Issue #756, opened 2018-02-06
The "Copy URL" icon and "Delete" button cannot be focused with "Tab" key


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 58.0.1 and up

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Prerequisites]
- Have one file uploaded to https://send.stage.mozaws.net/.

[Steps to reproduce]:
1. Start Firefox browser and navigate to "https://send.stage.mozaws.net/" page.
2. Focus the file from prerequisites with the "Tab" key.
3. Press "Tab" key once again and observe the behavior.

[Expected result]:
- The "Copy URL" icon is focused.

[Actual results]:
- The "Mozilla" footer link is focused.

[Notes]:
- This issue is also reproducible for the "Delete" button.
- This issue is also reproducible on Mac OS'es, but I've filed a separate issue (#755 ) because more clickable elements seem to be affected.
- Here is a screen recording with the issue:
![copy url and delete](https://user-images.githubusercontent.com/31244235/35861520-ac865334-0b50-11e8-9733-9a017f91b6ea.gif)

brainlulz:
I'm working on it :) we need to add tabindex

brainlulz:
Merge #840 close this issue, you can find it on the dev (https://send.dev.mozaws.net/).
Please close the issue if fixed. (Remind, look at my comment on Issue #755 for the Mac Os probem)

Softvision-CristinaBadescu:
Thanks for fixing this! The issue is no longer reproducible on https://send.dev.mozaws.net/, using the latest Nightly 62.0a1 (2018-06-06), on Windows 10 x64, Mac 10.13 and Arch Linux 4.12.

PR #840, opened 2018-06-03
fix: added a tabindex to the CopyUrl and Delete icon

brainlulz:
Added a tabindex to the "Copy URL" and "Delete" icon. This will make thoses icons focusable with Tab.

Fix issue #756


dannycoates:
Thank you @brainlulz 😄

PR #842, opened 2018-06-04
Added ECE transformers and tests

emily-hou1:
ECETransformer can be used to initialize TransformStreams that take a stream of plain data and output encrypted data, or vice versa.


app/blobslicer.js
```diff
@@ -0,0 +1,41 @@
...
+      let size = 0;
+      if (this.decrypt && this.index === 0) {
+        size = Math.min(21, bytesLeft);
+      } else {
+        size = Math.min(this.size, bytesLeft);
+      }
+      const blob = this.blob.slice(this.index, this.index + size);
+      const reader = new FileReader();
+      reader.onload = function() {
+        controller.enqueue(new Uint8Array(this.result));
```
martinthomson: You should use `() =>` rather than `function()` and use something other than `this`.

app/blobslicer.js
```diff
@@ -0,0 +1,41 @@
...
+  pull(controller) {
+    return new Promise((resolve, reject) => {
+      const bytesLeft = this.blob.size - this.index;
+      if (bytesLeft <= 0) {
+        controller.close();
+        return resolve();
+      }
+      let size = 0;
+      if (this.decrypt && this.index === 0) {
+        size = Math.min(21, bytesLeft);
```
martinthomson: This constant needs to be defined as one.

app/ece.js
```diff
@@ -0,0 +1,237 @@
...
+    header.rs = buffer.readUIntBE(KEY_LENGTH, 4);
+    const idlen = buffer.readUInt8(KEY_LENGTH + 4);
+    header.length = idlen + KEY_LENGTH + 5;
+    return header;
+  }
+
+  async encryptRecord(buffer, seq, isLast) {
+    const nonce = this.generateNonce(seq);
+    const encrypted = await window.crypto.subtle.encrypt(
+      { name: 'AES-GCM', iv: nonce },
```
martinthomson: You can probably define this struct as a constant and reuse it in a few places.

test/frontend/tests/streaming-tests.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    it('can encrypt', async function() {
+      const enc = new streams.TransformStream(
+        new ECETransformer('encrypt', key, rs, salt)
+      );
+
+      const rstream = await new BlobSliceStream(blob, rs - 17);
+
+      const reader = rstream.pipeThrough(enc).getReader();
```
martinthomson: Am I right in saying that the ECETransformer requires its input to be sliced into chunks that are exactly the right length?  That seems brittle.  Why not have it do its own slicing?

app/ece.js
```diff
@@ -0,0 +1,237 @@
...
+rs:   int containing record size, optional
+salt: ArrayBuffer containing salt of KEY_LENGTH length, optional
+The transform stream takes data as UInt8Arrays on the writable side, and outputs
+UInt8Arrays on the readable side.
+*/
+export default class ECETransformer {
+  constructor(mode, ikm, rs, salt) {
+    this.mode = mode;
+    this.prevChunk;
+    this.params = {};
```
martinthomson: Having a params struct rather than direct fields for key and nonce base is a little strange given the rest of this code.

app/ece.js
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  generateNonce(seq) {
+    const nonce = Buffer.from(this.params.nonceBase);
+    const m = nonce.readUIntBE(nonce.length - 4, 4);
+    const xor = (m ^ seq) >>> 0; //forces unsigned int xor
+    nonce.writeUIntBE(xor, nonce.length - 4, 4);
+
+    const m2 = nonce.readUIntBE(nonce.length - 8, 4);
+    const xor2 = (m2 ^ (seq >>> 4)) >>> 0;
```
martinthomson: IMPORTANT: you want to use a shift by 32, but that isn't going to work (bitwise ops in JS operate on 32-bit values).  Using floating point ops is necessary here, or you could decide not to support more than 2^32 records.

test/frontend/tests/streaming-tests.js
```diff
@@ -0,0 +1,88 @@
...
+      while (!state.done) {
+        result = decoder.decode(state.value);
+        state = await reader.read();
+      }
+
+      assert.equal(result, str);
+    });
+
+    it('can encrypt', async function() {
+      const enc = new streams.TransformStream(
```
dannycoates: we can improve the exposed interface of `ece.js` to do this stuff for us.

something like:

```js
const encryptedStream = new ECEEncryptStream(key, salt, rs, blob)
const reader = encryptedStream.getReader()

const decryptedStream = new ECEDecryptStream(key, blob)
```

PR #846, opened 2018-06-06
Fix #843: Upload image on paste

### LINKED ISSUE #843: Proposal: Upload image on paste
It is handy to auto upload image when user copy image and paste with `Cmd/Ctrl + V`.

Willing to submit a PR if it is accepted.

pd4d10:
Fix #843

### DIFF
--- a/app/pasteManager.js
+++ b/app/pasteManager.js
@@ -1,0 +1,24 @@
+/* global MAXFILESIZE */
+import { bytes } from './utils';
+
+export default function(state, emitter) {
+  window.addEventListener('paste', event => {
+    if (state.route !== '/' || state.uploading) return;
+
+    for (const item of event.clipboardData.items) {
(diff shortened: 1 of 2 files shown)


app/pasteManager.js
```diff
@@ -0,0 +1,24 @@
...
+
+      if (!file) continue; // Sometimes null
+
+      if (file.size > MAXFILESIZE) {
+        // eslint-disable-next-line no-alert
+        alert(state.translate('fileTooBig', { size: bytes(MAXFILESIZE) }));
+        continue;
+      }
+
+      emitter.emit('upload', { file, type: 'paste' });
```
dannycoates: We currently only allow one file to be uploaded at a time. How do you think we should handle that? One way would be to `return` here.

pd4d10: @dannycoates Nice catch. Updated

Issue #785, opened 2018-03-08
Implement streaming encryption


dannycoates:
Streams combined with RFC 8188 style encryption should allow us to support larger file sizes.

- https://developer.mozilla.org/en-US/docs/Web/API/Streams_API
- https://trac.tools.ietf.org/html/rfc8188
- https://github.com/jimmywarting/StreamSaver.js

dannycoates:
a stream that can slice blobs into chunks for encryption https://gist.github.com/dannycoates/99be55f18b0268241e01819f82f22a98

dannycoates:
done in vnext

hbakhtiyor:
@dannycoates did you consider about https://github.com/minio/sio/blob/master/DARE.md specification?

dannycoates:
no

PR #855, opened 2018-06-21
Streaming

emily-hou1:
Implements #785


emily-hou1:
#785

app/fileReceiver.js
```diff
@@ -51,6 +51,28 @@ export default class FileReceiver extends Nanobus {
...
+    let length = 0;
+
+    let state = await reader.read();
+    while (!state.done) {
+      chunks.push(state.value);
+      length += state.value.length;
+      state = await reader.read();
+    }
+
+    const result = new Int8Array(length);
```
dannycoates: Should this be an unsigned `UInt8Array`?

app/keychain.js
```diff
@@ -179,6 +179,16 @@ export default class Keychain {
     return ciphertext;
   }
 
+  async encryptStream(plaintext) {
```
dannycoates: do these need to be `async`?

app/ece.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+    if (salt === undefined) {
+      salt = generateSalt(KEY_LENGTH);
+    }
+
+    this.streamInfo = {
+      recordSize: rs,
+      fileSize: input.size + 16 * Math.floor(input.size / (rs - 17))
+    };
+    input = new BlobSliceStream(input, rs, mode);
```
dannycoates: It's usually easier to understand code when parameters aren't reassigned. Something like `const blobStream = ...` is more clear.

app/api.js
```diff
@@ -91,47 +91,102 @@ export async function setPassword(id, owner_token, keychain) {
...
+  const metadataHeader = arrayToB64(new Uint8Array(metadata));
+  const fileMeta = {
+    fileMetadata: metadataHeader,
+    authorization: `send-v1 ${verifierB64}`
+  };
+
+  function listenForResponse() {
+    return new Promise((resolve, reject) => {
+      ws.addEventListener('message', function(msg) {
+        const response = JSON.parse(msg.data);
```
dannycoates: We should catch parse errors and reject the promise, otherwise the promise may never get fullfilled.

app/api.js
```diff
@@ -91,47 +91,102 @@ export async function setPassword(id, owner_token, keychain) {
...
+            id: response.id,
+            ownerToken: response.owner
+          });
+        }
+      });
+    });
+  }
+
+  const resPromise = listenForResponse();
+  ws.send(JSON.stringify(fileMeta));
```
dannycoates: What happens if the server replies with an error? It looks like we still try to send the whole file. One way to stop early is to use a similar technique to the cancelled flag. Set an `errored` flag in `listenForResponse` and check it in the loop.

app/api.js
```diff
@@ -91,47 +91,102 @@ export async function setPassword(id, owner_token, keychain) {
...
+    ws.send(buf);
+
+    onprogress([Math.min(streamInfo.fileSize, size), streamInfo.fileSize]);
+    size += streamInfo.recordSize;
+    state = await reader.read();
+  }
+
+  const response = await resPromise;
+
+  ws.close();
```
dannycoates: We should always close the socket, even if we threw an error earlier.

app/api.js
```diff
@@ -91,47 +91,102 @@ export async function setPassword(id, owner_token, keychain) {
...
   verifierB64,
   keychain,
   onprogress
 ) {
-  const xhr = new XMLHttpRequest();
-  const upload = {
+  const host = window.location.hostname;
+  const port = window.location.port;
+  const protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+  const ws = await asyncInitWebSocket(`${protocol}//${host}:${port}/api/ws`);
```
dannycoates: I think it makes more sense to construct the socket in `upload`. That way the lifespan of `ws` is fully contained in that function.

server/routes/ws.js
```diff
@@ -0,0 +1,74 @@
...
+        const fileInfo = JSON.parse(message);
+        const metadata = fileInfo.fileMetadata;
+        const auth = fileInfo.authorization;
+
+        if (!metadata || !auth) {
+          ws.send(
+            JSON.stringify({
+              error: 400
+            })
+          );
```
dannycoates: Should we close the connection here?

server/routes/ws.js
```diff
@@ -0,0 +1,74 @@
...
+          auth: auth.split(' ')[1],
+          nonce: crypto.randomBytes(16).toString('base64')
+        };
+
+        const protocol = config.env === 'production' ? 'https' : req.protocol;
+        const url = `${protocol}://${req.get('host')}/download/${newId}/`;
+
+        const limiter = new Limiter(config.max_file_size);
+        fileStream = wsStream(ws, { binary: true }).pipe(limiter);
+        storage.set(newId, fileStream, meta);
```
dannycoates: should we `await` here before sending the reply?

app/fileReceiver.js
emily-hou1: Yes because I want to concat all of the stream output, but ArrayBuffer doesn't have methods for direct manipulation. I can refactor this to loop through only once though.

app/api.js
```diff
@@ -91,47 +91,115 @@ export async function setPassword(id, owner_token, keychain) {
...
+    });
+  });
+}
+
+async function upload(
+  stream,
+  streamInfo,
+  metadata,
+  verifierB64,
+  keychain,
```
dannycoates: keychain looks unused

server/routes/ws.js
```diff
@@ -0,0 +1,74 @@
...
+        const metadata = fileInfo.fileMetadata;
+        const auth = fileInfo.authorization;
+
+        if (!metadata || !auth) {
+          ws.send(
+            JSON.stringify({
+              error: 400
+            })
+          );
+          ws.close();
```
dannycoates: `return`?

server/routes/ws.js
dannycoates: If we `await` here, I think it solves the problem from yesterday in fileManager.js with the 'navigate' event.

dannycoates: Turns out, no, because the websocket stream doesn't end until it's closed.

PR #860, opened 2018-07-06
Hook up the android kotlin code to the send js code

fzzzy:
- Generate a bundle for android from android/android.js
- Temporarily remove -p from webpack since uglify chokes on things in android.js for some reason
- Encryption runs properly but then things go wrong when api.js tries to post the file to the server since the origin is null and the protocol is file


android/android.js
```diff
@@ -0,0 +1,159 @@
+/* global window, document, fetch */
+
+const MAXFILESIZE = 1024 * 1024 * 1024 * 2;
+
+const EventEmitter = require('events');
+const emitter = new EventEmitter();
+
+function dom(tagName, attributes, children = []) {
```
dannycoates: `choo/html` does pretty much exactly this. Is there a reason to not use it here?

app/api.js
```diff
@@ -137,7 +137,12 @@ async function upload(
   const host = window.location.hostname;
   const port = window.location.port;
   const protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
-  const ws = await asyncInitWebSocket(`${protocol}//${host}:${port}/api/ws`);
+  const endpoint =
+    window.location.protocol === 'file:'
+      ? 'wss://send2.dev.lcip.org/api/ws'
```
dannycoates: How should we do this for real? A build-time variable in webpack maybe?

app/fileManager.js
```diff
@@ -108,8 +108,10 @@ export default function(state, emitter) {
       if (cancelBtn) {
         cancelBtn.hidden = 'hidden';
       }
-      await delay(1000);
-      await fadeOut('.page');
+      if (document.querySelector('.page')) {
```
dannycoates: 👍

app/api.js
fzzzy: Good question. I created an issue to track this. https://github.com/mozilla/send/issues/869

We should do it however other similar configuration of this type is handled. A build-time variable in webpack seems like the right thing. Are there other things configured in this way already?

Issue #824, opened 2018-05-24
Implement Firefox Accounts Onboarding UI


johngruen:
This spec describes onboarding UI for FXA https://mozilla.github.io/testpilot-assets/Send/Send_Redesign/#artboard3

In order to close this issue, we need to demonstrate
1. A CTA banner that, when clicked, changes the Site UI to messaging about the benefits of FXA
2. A hinted avatar that, when clicked, does the same
3. A benefits view that allows the user to enter their email to trigger an FxA flow

PR #876, opened 2018-07-31
New UI

emily-hou1:
Implemented according to these mockups: [https://mozilla.github.io/testpilot-assets/Send/Send_Redesign/](url)


public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
 maxPasswordLength = Maximum password length: { $length }
 # A short status message shown when there was an error setting the password
 passwordSetError = This password could not be set
+pageHeaderCredits = from the makers of Firefox
```
flodolo: Why is this starting lowercase? It would definitely benefit from a note explaining it.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+addFilesButton = Add file(s)
```
flodolo: Can we use a proper plural form here? Or just "Add files"?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+uploadFilesButton = Send
+uploadFileProgress = Sending
+uploadDropDragMessage = Drop files here
+uploadDropClickMessage = or click to select a file
```
flodolo: Same as above. It would help to understand why this starts lowercase. Also, do we really need to split this sentence in two parts?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+addPasswordMessage = Protect with password
+addPasswordLabel = Password:
+copyUrlLabel = Copy and share this link:
+passwordReminder = don't forget the password too
```
flodolo: Always use proper apostrophes instead of straight single quotes: ’ vs ' (also in other strings)

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+signInPromoText = Sign In/Up!
+signInExplanation = It's free and gives you many more Send options
+signInLearnMore = Learn more!
+downloadProgressButton = Downloading... { $progress }
```
flodolo: Always use single unicode character for ellipsis: `…` vs `...`. Also, when adding a variable, it helps to add a comment explaining what's going to replace it with an example.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+downloadMessage2 = Firefox Send lets you share files with a safe, private, and encrypted link that automatically expires to ensure your stuff does not remain online forever.
```
flodolo: "stuff"?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+signInEmailEnter = Enter your Email
+emailEntryPlaceholder = Email
+signInContinueMessage = to continue to Firefox Send
```
flodolo: Again: why is this starting lowercase?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +126,30 @@ passwordIsSet = Password set
...
+signInContinueButton = Continue
+accountBenefitTitle = With a free Firefox Account with Send you can:
```
flodolo: That double "with" sounds a bit weird (alert: non native speaker)

dannycoates:
Thanks for the comments @flodolo :)

These strings are for the next version of Send and aren't final yet, which reminds me I have a question for you about that...

We're developing this in parallel with maintaining the current version (in a separate branch). When we are ready for localization (getting closer) how should we go about it?

We expect to have a public beta site near the end of Q3, so both versions will be live for a little while. We don't expect to be making any string changes to the current version, except maybe for linking to the beta.

flodolo:
> We're developing this in parallel with maintaining the current version (in a separate branch). When we are ready for localization (getting closer) how should we go about it?

I'd suggest to add new strings to the existing file, possibly group the soon-to-be-obsolete strings together, so it's easier to remove them whenever possible.

dannycoates:
@flodolo ok, so we can merge the new strings into master, where they'll be picked up by pontoon, then merge pontoon commits back into the vnext branch. Does that sound right?

I assume we'd merge master into vnext manually, so the current version dev site will pick up strings as usual however the vnext site won't automatically update. Is that ok?

flodolo:
> @flodolo ok, so we can merge the new strings into master, where they'll be picked up by pontoon, then merge pontoon commits back into the vnext branch. Does that sound right?

Yes

> I assume we'd merge master into vnext manually, so the current version dev site will pick up strings as usual however the vnext site won't automatically update. Is that ok?

I might need context: there's a different site for this version (vnext)?

dannycoates:
> I might need context: there's a different site for this version (vnext)?

Yes, for at least a month or so we plan on hosting both versions in parallel on different domains. Like  old.reddit.com and www.reddit.com.

flodolo:
> Yes, for at least a month or so we plan on hosting both versions in parallel on different domains. Like old.reddit.com and www.reddit.com.

OK. Non-automatic update for that would be OK.

emily-hou1:
resolves #824

PR #916, opened 2018-08-30
Implemented FxA

dannycoates:


public/locales/en-US/send.ftl
```diff
@@ -84,6 +84,8 @@ errorPageHeader = Something went wrong!
 errorPageMessage = There has been an error uploading the file.
 errorPageLink = Send another file
 fileTooBig = That file is too big to upload. It should be less than { $size }.
+tooManyFiles = Only { $count } files can be uploaded at a time.
```
flodolo: This needs to be a proper plural string.
https://projectfluent.org/fluent/guide/selectors.html

public/locales/en-US/send.ftl
```diff
@@ -84,6 +84,8 @@ errorPageHeader = Something went wrong!
...
+tooManyArchives = Only { $count } archives are allowed.
```
flodolo: Same here, need a proper plural

public/locales/en-US/send.ftl
dannycoates: there is no singular case. `$count` will always be > 10

dannycoates: I'll add a comment

flodolo: Please check [this document](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals), and look for example at how Russian or Arabic form plurals. Plurals is not about 1 vs !=1 like in English.

flodolo: If you only want to manage "more than one" in English, you can still do

```
tooManyFiles = 
    { $count ->
       *[other] Only { $count } files can be uploaded at a time.
    }
```

But at least that will trigger the plural UI for all locales in Pontoon.

Technically, with Fluent locales are free to add a plural form, even if English doesn't have one, but I prefer English to be explicit.

PR #918, opened 2018-09-06
 Implement wss preference url in html; Update to work with the tip of vnext branch; allow viewing the android ui from the webpack server

fzzzy:


app/api.js
```diff
@@ -1,6 +1,19 @@
 import { arrayToB64, b64ToArray, delay } from './utils';
 import { ECE_RECORD_SIZE } from './ece';
 
+let fileProtocolWSSUrl =
+  (localStorage && localStorage.getItem('wssURL')) ||
```
dannycoates: `localStorage` here breaks the service worker. a try/catch should fix it

fzzzy: thanks.

fzzzy:
Made the change, tests are green now, merging.

PR #952, opened 2018-09-27
Use FluentResource to parse and serialize FTL files server-side

### LINKED ISSUE #611: Avoid using MessageContext's internal APIs
Hi. Thanks for using fluent!

I noticed that the code in `build/fluent_loader` makes use of `ctx._messages` which is meant as an internal property. The proper way of adding translations to a `MessageContext` is via the `addMessages` method. It takes a string as the argument. What is the purpose of using `toJSON` here? Are you trying to inline parsed translations into a dist file?

https://github.com/mozilla/send/blob/9410defab6156d29bb705f73aefbd09ec588678e/build/fluent_loader.js#L23

stasm:
The `_messages` property of `FluentBundle` is considered internal and I'd like to suggest to move Send away from using it. I don't want to make any guarantees that this internal API won't change in a minor release of `fluent`. `FluentResource` is a new public API which can be used to parse, cache and serialize FTL files.

Fixes #611.

### DIFF
--- a/build/fluent_loader.js
+++ b/build/fluent_loader.js
@@ -1,16 +1,8 @@
-const { FluentBundle } = require('fluent');
+const { FluentResource } = require('fluent/compat');
 const fs = require('fs');
 
-function toJSON(map) {
-  return JSON.stringify(Array.from(map));
-}
-
-function merge(m1, m2) {
-  const result = new Map(m1);
-  for (const [k, v] of m2) {
-    result.set(k, v);
-  }
-  return result;
+function toJSON(resource) {
+  return JSON.stringify(Array.from(resource));
 }
 
 module.exports = function(source) {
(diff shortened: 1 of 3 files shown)


build/fluent_loader.js
```diff
@@ -1,16 +1,8 @@
-const { FluentBundle } = require('fluent');
+const { FluentResource } = require('fluent/compat');
 const fs = require('fs');
 
-function toJSON(map) {
-  return JSON.stringify(Array.from(map));
-}
-
-function merge(m1, m2) {
```
stasm: In Fluent, localization works best when languages are fully isolated. Rather than merging the messages, I suggest that for non en-US locales, two `FluentResource` are included in the dist file. This ensures that inerpolations of strings within other strings are consistently translated, as well as that the formatting of numbers and dates corresponds to the language of the text around them. I implemented this approach in this PR.

build/fluent_loader.js
```diff
@@ -20,34 +12,45 @@ module.exports = function(source) {
...
-  context.addMessages(source);
 
-  const merged = merge(en._messages, context._messages);
+  // Parse the current language's translation file.
+  const locResource = FluentResource.fromString(source);
+  let enResource;
+
+  // If the current language is not en-US, also parse en-US to provide a
+  // fallback for missing translations.
+  if (locale !== 'en-US') {
```
stasm: Ideally, we should work witn an array of languages negotatied by `fluent-langneg`. I found `server/middleware/language.js` but I wasn't sure if changing `req.language` to an array was a good first step. I can look into it a bit more, if you'd like.

build/fluent_loader.js
```diff
@@ -20,34 +12,45 @@ module.exports = function(source) {
...
+  });
+
   function translate(id, data) {
-    var msg = bundle.getMessage(id);
-    if (typeof(msg) !== 'string' && !msg.val && msg.attrs) {
-      msg = msg.attrs.title || msg.attrs.alt
+    for (let bundle of bundles) {
+      if (bundle.hasMessage(id)) {
+        let message = bundle.getMessage(id);
+        return bundle.format(message, data);
```
stasm: I didn't see any messages in the current FTL file which use `title` or `alt`. We're going to change the `format` API soon to allow formatting attributes, too (probably via `format("string-id.attribute-name")`). I only implemented formatting message values here. If formatting attributes is required now, I can add it in this PR. Otherwise, I suggest to wait for the next release of `fluent`.

stasm:
@dannycoates Are you the right person to review this?

dannycoates:
Thanks @stasm! I'll review this today

dannycoates:
Thanks @stasm! I'll review this today

build/fluent_loader.js
dannycoates: we don't use any attributes atm so 👍

build/fluent_loader.js
dannycoates: I created a follow up issue to dig into this, #955. No need to hold up merging this in the meantime :)

dannycoates:
Looks great! 💯

PR #954, opened 2018-09-27
Initial commit switching current integration tests to javascript.

jrbenny35:
As per title. This change wasn't really needed but with the addition of the react-native apps, testing those with [detox](https://github.com/wix/detox) requires javascript, so I don't see the need for having 2 different languages for the same type of tests.

This PR should include all of the current python/pytest tests that we have as well as an additional download test files.


jrbenny35:
@dannycoates I think this is good for a look.

dannycoates:
This looks great @jrbenny35. I would like to retarget this to the vnext branch since it'll be merged back to master soon. The only part that I can tell will have conflicts is the dependencies. If you'd rather not do it I can give it a shot tomorrow.

jrbenny35:
@dannycoates is the UI different for that branch?

dannycoates:
> is the UI different for that branch?

yes

jrbenny35:
I'll get this updated to work on vnext, if you could handle the dependencies tomorrow, that would be awesome.

jrbenny35:
@dannycoates this has been updated and tested against https://vnext.send.nonprod.cloudops.mozgcp.net/. The circle job won't pass but I can confirm it works. If something breaks after merging I can fix it on vnext

dannycoates:
closing in favor of #960

PR #956, opened 2018-09-28
update ui

johngruen:
@dannycoates this is a partial ui update. A few notes:

* definitely breaks a few things, particularly responsive views
* I think we'll need to change things around so there's not as much stuff in the index route template and more stuff in the welcome page. I can file separate bugs for this stuff once mark and I finalize UI.


public/locales/en-US/send.ftl
```diff
@@ -152,8 +152,8 @@ addPasswordMessage = Protect with password
 addPasswordLabel = Password:
 copyUrlLabel = Copy and share this link:
 passwordReminder = don't forget the password too
-signInPromoText = Sign In/Up!
-signInExplanation = It's free and gives you many more Send options
+signInPromoText = Sign In/Up
+signInExplanation = It's free and you can send bigger files.
```
flodolo: Need new string IDs for both messages
https://developer.mozilla.org/en-US/docs/Making_String_Changes

dannycoates: these strings haven't hit master yet

app/pages/welcome/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = function(state, emit) {
   }
 
   return html`
-  <div id="page-one" class="page ${fade}">
```
dannycoates: 🍺 pour one out for page-one, the og div.

PR #960, opened 2018-10-03
Vnext integration tests

dannycoates:


jrbenny35:
The tests are failing because we need to run them either with a local firefox or with docker-compose, like how the other PR was running. If we can do a ```docker-compose up -d``` and then ```docker-compose exec selenium-firefox npm run test-circle:selenium```. For that we will need to switch back to machine. And we will have to update the node-version within the selenium-firefox image.

PR #961, opened 2018-10-04
gitignore update- add lock file

maddhruv:


dannycoates:
I'd prefer to leave editor specific ignores out of our .gitignore. I recommend adding this globally to your environment. For example: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #966, opened 2018-10-09
Make file download limit changeable in config

jrmedd:
We're testing using Send internally, and needed to customise a few parameters to make it more suitable: 

Finding ```expire_config``` in **config.js** was logical and easy, but increasing the download limit options meant going to **app/pages/share/index.js** to change the options variable, and then **server/routes/params.js** to make the server accept anything above 20. 

**N.B. I'm really not much of a Node guy, so I expect some tests might not work**

### DIFF
--- a/app/pages/share/index.js
+++ b/app/pages/share/index.js
@@ -1,4 +1,6 @@
 /* global EXPIRE_SECONDS */
+/* global EXPIRE_LIMIT */
+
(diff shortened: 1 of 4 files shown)


dannycoates:
Thanks for the PR @jrmedd. We recently merged a big change that includes this functionality. The environment variable is `DOWNLOAD_COUNTS`.

PR #970, opened 2018-10-15
Fix: the font-weight is too thin to readable

zcyzcy88:
Current font-weight is terrible and should be changed to the same as https://testpilot.firefox.com/ and https://www.mozilla.org/.


dannycoates:
Thanks for the PR @zcyzcy88. We've completely changed the UI so I'm closing this now.

Issue #976, opened 2018-10-24
Production build not working


hhoechtl:
If you try to install only production dependencies and build the production frontend it doesn't work atm.

```shell
➜ npm install --production
➜ npm run build
npm WARN using --force I sure hope you know what you are doing.

> firefox-send@2.5.5 build /app
> npm run clean && webpack -p


> firefox-send@2.5.5 clean /app
> rimraf dist

sh: rimraf: not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! firefox-send@2.5.5 clean: `rimraf dist`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the firefox-send@2.5.5 clean script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Only installing the dev dependencies fixes this issue.

Also the `test/frontend/index.js` is always needed for a webpack build (which shouldn't for production) => relevant for docker build as it should be in .dockerignore

And another one: `git-rev-sync` requires the projectroot to be a git repository. But I don't want to keep the .git files in a docker build.

dannycoates:
Hi @hhoechtl :)

Our dockerfile is designed to be run *after* an `npm run build` on the dev machine. This saves us from installing build-time dependencies in the container. See https://github.com/mozilla/send/blob/master/docs/docker.md

hhoechtl:
Thx for clarification

PR #977, opened 2018-10-24
Fix and optimize docker build

hhoechtl:
Makes docker build work temporarily. But there are other issues to address, see #976 .


dannycoates:
https://github.com/mozilla/send/issues/976#issuecomment-432882940

PR #998, opened 2018-11-06
Changed sign in area

### LINKED ISSUE #989: Make CTA more clear in vnext design
The little shilouette head ain't doing much for me...could we do something a little bit more clear here?

Maybe just a blue button that says sign in/up?

screen shot 2018-11-04 at 7 25 07 pm

dannycoates:
How about?
screen shot 2018-11-06 at 2 43 14 pm
screen shot 2018-11-06 at 2 42 52 pm

fixes #989


youwenliang:
I would suggest something like this to make to button more like a button:
image

PR #1078, opened 2019-01-09
[Integration-Test] Add test for number of downloads, fixes #810

### LINKED ISSUE #810: [Integration-Test] Add test for number of downloads.
Per testrail:

Send - Numbers of Downloads - 2 

1. Click the  "1 download" button and select the "2" option. | The "1 download" button changes into a "2 downloads" button.
2. Navigate to the Send homepage and observe the numbers of downloads for the uploaded file. | The 0/2 (0 downloads out of two) is correctly displayed.
3. Click the the uploaded file name. | You are redirected to the Share links page. The "2 downloads" button is correctly displayed.

We should add to the option to configure downloads to the ```share.py``` Page Object Model. The test should check 1. After 1 download, the remaining downloads left are 1 less than they were before. 2. The homepage correctly shows the amount of downloads left.

dhyey35:
Added a test for `2 downloads` being changed to `1 download` on the home page.

### DIFF
--- a/app/ui/expiryOptions.js
+++ b/app/ui/expiryOptions.js
@@ -42,7 +42,8 @@ module.exports = function(state, emit) {
         }
         state.downloadCount = value;
         emit('render');
-      }
+      },
+      'expire-after-dl-count-dropdown'
     ),
     dlCountSelect
   );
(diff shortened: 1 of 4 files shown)


dannycoates:
Thank you @dhyey35, this looks great! I'm going to do a follow up commit to change the `expiresAfterText` selector to something more direct. It works fine as is, I just don't want to rely on style classes for the lookup.

PR #1086, opened 2019-01-10
Make ANDROID=1 npm start work again, as it broke at some point

fzzzy:


server/bin/dev.js
```diff
@@ -30,7 +33,7 @@ module.exports = function(app, devServer) {
     app.get('/', android);
     app.get('/legal', android);
     app.get(`/share/:id${ID_REGEX}`, android);
-    app.get(`/download/:id${ID_REGEX}`, android);
+    app.get(`/download/:id${ID_REGEX}/:key`, android);
```
dannycoates: this needs to go on the choo route in `android.js` not the express route

fzzzy: Ok sure that makes sense... But this fixes the bug???

dannycoates: nevermind, i was confused about what we're trying to do here

PR #1087, opened 2019-01-11
use custom elements instead of adding css classes for selectors

dannycoates:
I'd like to try keeping our use of `class` for style only with no semantic meaning implied, but we need a nice way to select ui elements in tests/etc. I propose using custom elements (using `send-` as the prefix). See how I've modified  `download-tests.js` to use the element as an example.

I probably didn't rename everything that we want to classify, so we can change them as we go. I'd prefer not to make *everything* a custom element so if the selector to get to a specific element within a custom element isn't too complicated I'd prefer that over making new elements. For example the `'send-archive > div'` selector in the download test.

@fzzzy @jrbenny35 thoughts?

### DIFF
--- a/app/ui/account.js
+++ b/app/ui/account.js
@@ -62,18 +62,18 @@ class Account extends Component {
     const translate = this.state.translate;
     if (!this.local.loggedIn) {
       return html`
-        <div>
+        <send-account>
           <button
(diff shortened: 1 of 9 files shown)


jrbenny35:
This is pretty great! I like it and I think its a simple change that will aid contributors and future tests.

jrbenny35:
Would this be extended to the android implementation as well? That would be great, as those tests will be quite similar.

PR #1091, opened 2019-01-11
add new metrics md

johngruen:


johngruen:
Pulling comments from the gist made by @irrationalagent

Generally the taxonomy LGTM, though obviously I don't have enough context on how send works to know if some things are technically possible, such as keeping the a running total of those user properties (number of downloads, etc) and the feasibility of fetching them for every applicable event.

some questions/comments:

    Is the device_id always specific to send? If you could you use the fxa device_id iif a user has an account it could help in the future to join on fxa telemetry. but if you aren't caching the FxA profile data anywhere in the app it might be hard to do.
    fxa has a user property called entrypoint - it might be useful for you to add this or something similar here if there are many places that a user could enter the singup/reg funnel from (i.e. if there are multiple forms or links that lead to registration/signin). Please ignore if this is the same as FxA prompt trigger
    keep in mind that running-total type of user properties in amplitude will reflect the value of that property at the time of the events/periods you're looking at. so if you make a cohort of "users that had 0 uploads in the past 30 days" it will pick out users that had 0 uploads at any point in the last 30 days - they could have had more than 0 as well. there are ways to get "users with 0 uploads and only 0", but its just something to be aware of.
  

Also let me know if you need help setting things up so that you send us the proper metrics parameters when someone begins the fxa login/reg flow.

PR #1099, opened 2019-01-16
Fix #888 Implement share card.

### LINKED ISSUE #888: Implement Share Cards in Send Android
Section 6.1

fzzzy:


android/app/src/main/java/org/mozilla/sendandroid/MainActivity.kt
```diff
@@ -35,6 +35,11 @@ class WebAppInterface(private val mContext: MainActivity) {
     fun beginOAuthFlow() {
         mContext.beginOAuthFlow();
     }
+
+    @JavascriptInterface
+    fun shareUrl(url: String) {
+        mContext.shareUrl(url)
```
dannycoates: off topic... is there a kotlin linter/formatter we should use?

app/ui/archiveTile.js
```diff
@@ -140,6 +142,26 @@ function archiveDetails(translate, archive) {
 }
 
 module.exports = function(state, emit, archive) {
+  const copyOrShare =
+    typeof window.Android !== 'object'
```
dannycoates: we should standardize how we check for android. I used `utils/browserName` somewhere else to do the same thing. I don't really care how we do it but we should probably wrap it in a function.

android/app/src/main/java/org/mozilla/sendandroid/MainActivity.kt
fzzzy: I'll ask the lockbox team.

app/ui/archiveTile.js
fzzzy: Good call. I switched to browserName.

PR #1125, opened 2019-01-23
added download link to tile

### LINKED ISSUE #1081: Add a download button to upload bundle cards
Supports first party downloading for fxa connected users with multiple devices

dannycoates:
still needs the icon

fixes #1081

### DIFF
--- a/app/ui/archiveTile.js
+++ b/app/ui/archiveTile.js
@@ -4,11 +4,11 @@ const html = require('choo/html');
 const raw = require('choo/html/raw');
 const assets = require('../../common/assets');
 const {
-  browserName,
   bytes,
   copyToClipboard,
   list,
   percent,
+  platform,
   timeLeft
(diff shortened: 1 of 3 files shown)


app/ui/archiveTile.js
```diff
@@ -169,6 +164,19 @@ module.exports = function(state, emit, archive) {
             <img src="${assets.get('share-24.svg')}" class="mr-2" /> Share
           </button>
         `;
+  const dl =
+    platform() === 'web'
```
fzzzy: I think the download link would be just as useful on mobile. A good way to get a file from a desktop machine to a mobile device?

dannycoates: interesting. should we add the download ui to the android app then?

fzzzy: It could just be a link to the server.

PR #1128, opened 2019-01-30
fix 1112

### LINKED ISSUE #1112: Some layout issues with large format layouts for android
Might effect tablets

screen shot 2019-01-22 at 9 16 04 am
screen shot 2019-01-22 at 9 16 14 am

youwenliang:
fix #1112

### DIFF
--- a/app/ui/archiveTile.js
+++ b/app/ui/archiveTile.js
@@ -29,7 +29,7 @@ function password(state) {
   const MAX_LENGTH = 32;
 
   return html`
-    <div class="my-2 px-4 md:px-0">
+    <div class="my-2 px-4">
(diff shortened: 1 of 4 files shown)


dannycoates:
off topic but... package-lock is so annoying. i think we should start to only commit changes to it when we also have a package.json change (like an `npm update`). cc @fzzzy

`git checkout -- package-lock.json` resets it after npm does it's bs.

fzzzy:
I agree about package-lock. Recently I have just not been adding it to my commits. I think we should update it whenever package.json changes, and also maybe the dependency update bot can update it? Assuming we actually merge those dependency updates

PR #1134, opened 2019-01-31
gcs: Increase timeouts from 60s to 86400s

jbuck:
No idea whether this will actually work or not, but based on comments [in this issue](https://github.com/googleapis/nodejs-storage/issues/27#issuecomment-435252701) I [increased fetch timeouts from 60s to 86400s](https://github.com/googleapis/nodejs-storage/compare/master...jbuck:master).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -943,9 +943,9 @@
       }
     },
     "@google-cloud/common": {
-      "version": "0.27.0",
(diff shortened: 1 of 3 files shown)


dannycoates:
Thanks @jbuck, for finding the root cause. Instead of this patch I've reverted `@google-cloud/storage` to 2.1.0 which doesn't have this issue. I'll file a follow up issue to investigate how to move forward, but we're unblocked now.

PR #1141, opened 2019-02-07
implemented amplitude metrics

dannycoates:


app/archive.js
```diff
@@ -17,6 +17,9 @@ function isDupe(newFile, array) {
 export default class Archive {
   constructor(files = []) {
     this.files = Array.from(files);
+    this.timeLimit = DEFAULTS.EXPIRE_SECONDS;
+    this.dlimit = 1;
+    this.password = null;
```
dannycoates: I moved these fields in from choo `state`. It made the metric call easier with the data in one place, and it seems like a better place anyway

app/controller.js
```diff
@@ -50,37 +50,27 @@ export default function(state, emitter) {
 
   emitter.on('logout', () => {
     state.user.logout();
-    state.timeLimit = DEFAULTS.EXPIRE_SECONDS;
-    state.downloadCount = 1;
+    metrics.loggedOut({ trigger: 'button' });
     emitter.emit('pushState', '/');
   });
 
-  emitter.on('changeLimit', async ({ file, value }) => {
```
dannycoates: this event was orphaned with the new ui

app/controller.js
```diff
@@ -100,20 +90,35 @@ export default function(state, emitter) {
...
+          size: bytes(maxSize),
+          count: LIMITS.MAX_FILES_PER_ARCHIVE
+        })
+      );
     }
     render();
   });
 
-  emitter.on('upload', async ({ type, dlimit, password }) => {
+  emitter.on('signup-cta', source => {
```
dannycoates: new event to simplify creating the dialog from multiple places and tracking the source

app/controller.js
```diff
@@ -100,20 +90,35 @@ export default function(state, emitter) {
...
+    state.modal = signupDialog(source);
+    render();
+  });
+
+  emitter.on('authenticate', async (code, oauthState) => {
```
dannycoates: moved here from `routes.js`. it was metrics related for a minute but then i got rid of that metric. seems better here anyway

app/ownedFile.js
```diff
@@ -8,7 +8,6 @@ export default class OwnedFile {
     this.url = obj.url;
     this.name = obj.name;
     this.size = obj.size;
-    this.type = obj.type;
```
dannycoates: cryptically this was related to old metrics. `type` was 'click' or 'drag'. no longer used

PR #1146, opened 2019-02-08
npm audit fix

jbuck:


jbuck:
@dannycoates v2.6.2 was failing CI because of a security vulnerability - this is the result of running `npm audit fix`

PR #1147, opened 2019-02-09
Fix integration tests for multiple files

dhyey35:
According to @jrbenny35 's comment [here](https://github.com/mozilla/send/issues/810#issuecomment-460341441) I have added a new class for uploaded files. 

I have tried according to my understanding but I am not able to figure out a good way to get the `archiveId` without using [waitUntil](http://v4.webdriver.io/api/utility/waitUntil.html) but that make the code messy. So I thought I should submit my work in progress code to get help.

@jrbenny35 Can you please tell me if I am going in right direction or if another approach is required

### DIFF
--- a/test/integration/download-tests.js
+++ b/test/integration/download-tests.js
@@ -42,17 +42,23 @@ describe('Firefox Send', function() {
     );
     browser.waitForExist(homePage.uploadButton);
     browser.waitForExist(homePage.downloadCountSelect);
+    // Select 2 downloads
     browser.selectByIndex(homePage.downloadCountSelect, 1);
     browser.click(homePage.uploadButton);
+    // TODO: need something like `await browser.waitForExist(homePage.shareUrl)`
     browser.waitForExist(homePage.shareUrl);
+    // Newly added file will always be first in the list
+    const archiveId = browser.getAttribute('send-archive', 'id')[0];
     const downloadPage = new DownloadPage(browser.getValue(homePage.shareUrl));
+    downloadPage.setUseCase(archiveId, 'CHECK_DL_COUNT');
     downloadPage.open();
     downloadPage.download();
     browser.waitForExist(downloadPage.downloadComplete);
     browser.back();
-    browser.waitForExist('send-archive');
+    browser.waitForExist(`#${archiveId}`);
+    const uploadedFile = downloadPage.getUploadedFiles('CHECK_DL_COUNT')[0];
     assert.equal(
-      browser.getText('send-archive > div').substring(0, 24),
+      uploadedFile.getExpiryText().substring(0, 24),
       'Expires after 1 download'
     );
   });
--- a/test/integration/pages/desktop/download_page.js
+++ b/test/integration/pages/desktop/download_page.js
@@ -1,6 +1,12 @@
 /* global browser */
 const Page = require('./page');
 
+/*
+  An object with archiveId as key and useCase as value
+  Can be used to retrive files according to it's use case
+*/
+const uploadedFilesUseCase = {};
+
 class DownloadPage extends Page {
   constructor(path) {
     super(path);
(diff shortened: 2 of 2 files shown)


jrbenny35:
Hey @dhyey35, I will take a look at this tomorrow but so far it looks good.

test/integration/download-tests.js
```diff
@@ -42,17 +42,23 @@ describe('Firefox Send', function() {
...
+    downloadPage.setUseCase(archiveId, 'CHECK_DL_COUNT');
     downloadPage.open();
     downloadPage.download();
     browser.waitForExist(downloadPage.downloadComplete);
     browser.back();
-    browser.waitForExist('send-archive');
+    browser.waitForExist(`#${archiveId}`);
+    const uploadedFile = downloadPage.getUploadedFiles('CHECK_DL_COUNT')[0];
     assert.equal(
-      browser.getText('send-archive > div').substring(0, 24),
```
dannycoates: This selector broke when we added another child div for the download link. All that needs to happen is change it to `send-archive > div:first-of-type`. If you'd rather not do it here we could also add it as another property in `pages/desktop/download_page.js`

jrbenny35:
https://github.com/mozilla/send/issues/810 this is the discussion that led to this PR.

dannycoates:
Yes, I read #810 before commenting. I understand the problem, it's this solution that I don't. The `useCase` concept is confusing and doesn't seem necessary, `UploadedFile` shouldn't be in  `download_page.js` (probably `home_page.js`), and `getUploadedFiles` should probably get populated from a selector. 

The idea is fine, just the implementation needs work. A better solution will probably present itself once the code is in the right place, `home_page.js`.

jrbenny35:
> `UploadedFile` shouldn't be in `download_page.js` (probably `home_page.js`), and `getUploadedFiles` should probably get populated from a selector.

I agree.

test/integration/pages/desktop/download_page.js
```diff
@@ -23,5 +29,41 @@ class DownloadPage extends Page {
   download() {
     return browser.click(this.downloadButton);
   }
+
+  setUseCase(id, useCase) {
```
jrbenny35: Could you explain your thought process behind this? I am having a hard time seeing exactly what it does.

dhyey35: I thought that we would have an object which keeps track of all the uploaded files by the tests so other tests can also use them in future. So I created `uploadedFilesUseCase` private object which stores `{ 'archive-#id': 'some_use_case' } `  key-value pairs. That's why we have `setUseCase` and `getUploadedFiles` methods.

For example let's say one test uploaded two files called `file1.txt` and `file2.txt` for testing if uploads work and set's it's USE_CASE to `UPLOAD_TESTING` and we then ran a couple of tests in between so `file1.txt` and `file2.txt` were no longer the topmost files in DOM then there was another test case which checked if downloading works fine, so instead of uploading new files again it can get the two files used for upload testing by calling `getUploadedFiles('UPLOAD_TESTING')` and then it can locate the two files by querying DOM with the archive-id's returned by the `getUploadedFiles` function.

I took some inspiration from the [link that you mentioned](https://github.com/mozilla/notes/blob/master/test/integration/page_objects/list_page.js#L50) and created `UploadedFile` class.

But now I see that this creates a relationship between tests ( one test depends on another - change in one requires change in another too - which is bad ) 

Now I think as @dannycoates mentioned this was an overkill.

dhyey35: > Yes, I read #810 before commenting. I understand the problem, it's this solution that I don't. The `useCase` concept is confusing and doesn't seem necessary, `UploadedFile` shouldn't be in `download_page.js` (probably `home_page.js`), and `getUploadedFiles` should probably get populated from a selector.
> 
> The idea is fine, just the implementation needs work. A better solution will probably present itself once the code is in the right place, `home_page.js`.

Please let me know if I should start working in this direction.

jrbenny35: You have the right idea but an object isn't needed. My link was just the object to represent the items on the page. If you see [this](https://github.com/mozilla/notes/blob/master/test/integration/page_objects/list_page.js#L26) line, you will see that we create a new list of objects and return it each time that function is called. 

This is how we can get around having a separate object that must keep track of the elements. Each time we call that function it will get and create a list of objects representing what is on the page at that specific moment. 

Let me know if that helps.

clouserw:
I'm going to close this due to bitrot.  If there is an idea to pursue here please get in touch with us and we'll get a new patch going.  Thanks

PR #1167, opened 2019-02-22
finish string update

johngruen:


public/locales/en-US/send.ftl
```diff
@@ -15,7 +15,7 @@ importingFile = Importing…
 verifyingFile = Verifying…
 encryptingFile = Encrypting…
 decryptingFile = Decrypting…
-notifyUploadEncryptDone = Your file is encrypted and ready to send.
```
johngruen: This is a new string

PR #1175, opened 2019-02-28
added warning page for browsers without streams downloading big files

dannycoates:
On browsers that don't fully support streams (like Safari & Edge) with a download > 256MB you'll get a screen like this:
screen shot 2019-02-28 at 2 12 39 pm

The button text and action changes based on which option is select

screen shot 2019-02-27 at 7 57 04 pm
screen shot 2019-02-27 at 7 57 15 pm


dannycoates:
strings aren't final

PR #1182, opened 2019-03-02
1149-fix-aborted-event-gt-node-v9.x.x

### LINKED ISSUE #1149: This link has expired or never existed in the first place!
Hi,

I am sorry if it not a right place to write about it.
I was trying to use the send secure tool, but I have problem with uploading the file.
It always redirecting me to notFound page, error in browser : "POST http://localhost:1443/api/info/fc6f2ba2a4 404 (Not Found)"
After some hustle with code (changing port,changing environment-> it would be great if you will update the documentation) I have still had my issue.
Could you gave me some direction, tips what it could be? I don't exactly understand the mechanism, that it stores the file in tmp directory, but for me it is not storing anything, just creating directory, so probably this is the reason that the program cannot find the uploaded filed. 

I would appreciate any help and guide.

Cheers,
Karolina

jowy:
Resolves #1149.

Use `aborted` rather than `close` event for detecting when the client has cancelled the upload or download requests.

Note: Broken with node `> v11.0.0`, resolved in node `v11.10.1`

### DIFF
--- a/server/routes/download.js
+++ b/server/routes/download.js
@@ -9,7 +9,7 @@ module.exports = async function(req, res) {
     const file_stream = storage.get(id);
     let cancelled = false;
 
-    req.on('close', () => {
+    req.on('aborted', () => {
       cancelled = true;
       file_stream.destroy();
     });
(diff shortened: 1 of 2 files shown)


dannycoates:
Thanks for the PR @jowy, I'll be able to look at it tomorrow.

dannycoates:
vnext got merged into master so this is now obsolete

PR #1183, opened 2019-03-03
fixed the port listed in the readme for the dev ui

nickmknia:
when running the development environment following the steps in the README, node is listening on port 1443.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ npm install
 npm start
 ```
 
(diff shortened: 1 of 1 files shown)


dannycoates:
Thanks for the PR @nickmknia. We recently merged a big change that happened to correct the problem in the other direction, `npm start` now uses port 8080 again.

Issue #1097, opened 2019-01-16
[Send V2] Download link expires also if you cancel the download before completion


poubina:
[Affected Versions]:

Firefox 64 and up

[Affected Platforms]:

All Windows
All Mac
All Linux

[Steps to reproduce]:

1. Open the Firefox browser and navigate to https://send2.dev.lcip.org/.
2. Click the "Add file(s)" button.
3. Select a file and click the "Open" button.
4. Set "Expires after" option to 2 downloads (*).
5. Upload the file.
6. Copy the link to clipboard.
7. On a new tab of the browser, copy and paste the link of previous step.
8. Download and CANCEL twice (*) the downloading before completion. 
9. Observe the behavior.

[Expected result]:

Download link must not expire after 2 tries, as we are cancelling the downloading before it gets completed.

[Actual result]:

Download link expires after cancelling it for the second time.

[Notes]:

Please find screenshots attached:

![cancel1](https://user-images.githubusercontent.com/44272078/51250657-d3fdbb80-1975-11e9-8c79-b5d79a884772.png)

![cancel2](https://user-images.githubusercontent.com/44272078/51250666-dbbd6000-1975-11e9-8b3c-c69564d38d9b.png)

jowy:
Should be resolved by PR https://github.com/mozilla/send/pull/1184

PR #1184, opened 2019-03-03
1149-vnext-fix-aborted-event-gt-node-v9.x.x

jowy:
Related to Issue https://github.com/mozilla/send/issues/1149

See PR https://github.com/mozilla/send/pull/1182


jowy:
Solves #1097 as well.

dannycoates:
the socket events we see trigger are a little different when run behind our load balancer. i need to do some more work get this sorted out there.

PR #1188, opened 2019-03-04
added vnext strings for translation

dannycoates:
Strings shared between prod and vnext are added as comment lines so that once prod strings are no longer needed we can delete all lines above and uncomment shared. vnext only strings are appended to the bottom.

### DIFF
--- a/public/locales/en-US/send.ftl
+++ b/public/locales/en-US/send.ftl
@@ -117,3 +117,146 @@ passwordIsSet = Password set
 maxPasswordLength = Maximum password length: { $length }
 # A short status message shown when there was an error setting the password
 passwordSetError = This password could not be set
+#
+#
+#
+# shared ->
+#
+#
+#
+#  title = Firefox Send
+#  siteFeedback = Feedback
+#  importingFile = Importing…
+#  encryptingFile = Encrypting…
+#  decryptingFile = Decrypting…
+#  downloadCount = { $num ->
+#          [one] 1 download
+#         *[other] { $num } downloads
+#      }
+#  timespanHours = { $num ->
+#          [one] 1 hour
+#         *[other] { $num } hours
+#      }
+#  copiedUrl = Copied!
+#  unlockInputPlaceholder = Password
+#  unlockButtonLabel = Unlock
+#  downloadButtonLabel = Download
+#  downloadFinish = Download complete
+#  fileSizeProgress = ({ $partialSize } of { $totalSize })
+#  sendYourFilesLink = Try Firefox Send
+#  errorPageHeader = Something went wrong!
+#  fileTooBig = That file is too big to upload. It should be less than { $size }.
+#  linkExpiredAlt = Link expired
+#  notSupportedHeader = Your browser is not supported.
+#  notSupportedLink = Why is my browser not supported?
+#  notSupportedOutdatedDetail = Unfortunately this version of Firefox does not support the web technology that powers Firefox Send. You’ll need to update your browser.
+#  updateFirefox = Update Firefox
+#  deletePopupCancel = Cancel
+#  deleteButtonHover = Delete
+#  footerLinkLegal = Legal
+#  footerLinkPrivacy = Privacy
+#  footerLinkCookies = Cookies
+#  passwordTryAgain = Incorrect password. Try again.
+#  javascriptRequired = Firefox Send requires JavaScript
+#  whyJavascript = Why does Firefox Send require JavaScript?
+#  enableJavascript = Please enable JavaScript and try again.
+#  # A short representation of a countdown timer containing the number of hours and minutes remaining as digits, example "13h 47m"
+#  expiresHoursMinutes = { $hours }h { $minutes }m
+#  # A short representation of a countdown timer containing the number of minutes remaining as digits, example "56m"
+#  expiresMinutes = { $minutes }m
+#  # A short status message shown when the user enters a long password
+#  maxPasswordLength = Maximum password length: { $length }
+#  # A short status message shown when there was an error setting the password
+#  passwordSetError = This password could not be set
+#
+#
+#
+# vnext ->
+#
+#
+#
+introTitle = Simple, private file sharing
+introDescription = Firefox Send lets you share files with end-to-end encryption and a link that automatically expires. So you can keep what you share private and make sure your stuff doesn't stay online forever.
+notifyUploadEncryptDone = Your file is encrypted and ready to send
+archiveExpiryInfo = Expires after { $downloadCount } or { $timespan }
+timespanMinutes = { $num ->
+        [one] 1 minute
+       *[other] { $num } minutes
+    }
+timespanDays = { $num ->
+        [one] 1 day
+       *[other] { $num } days
+    }
+timespanWeeks = { $num ->
+        [one] 1 week
+       *[other] { $num } weeks
+    }
+fileCount = { $num ->
+    [one] 1 file
+   *[other] { $num } files
+}
+totalSize = Total size: { $size }
+copyLinkDescription = Copy the link to share your file:
+copyLinkButton = Copy link
+downloadTitle = Download files
+# Firefox Send is a brand name and should not be localized.
+downloadDescription = This file was shared via Firefox Send with end-to-end encryption and a link that automatically expires.
+trySendDescription = Try Firefox Send for simple, safe file sharing.
+# count will always be > 10
+tooManyFiles = { $count ->
+    *[other] Only { $count } files can be uploaded at a time.
+}
+# count will always be > 10
+tooManyArchives = { $count ->
+    *[other] Only { $count } archives are allowed.
+}
+expiredTitle = This link has expired.
+# Firefox Send is a brand name and should not be localized.
+notSupportedDescription = Firefox Send will not work with this browser. Send works best with the latest version of Firefox, and will work with the current version of most browsers.
+downloadFirefox = Download Firefox
+legalTitle = Send Privacy Notice
+legalDateStamp = Version 1.0, dated March 12, 2019
+legalNoticeSend = When Mozilla receives information from you, our <a>Mozilla Privacy Policy</a> describes how we handle that information. Below are the top things you should know about Firefox Send.  You can also view the code <a>here</a>.
+legalNoticeSendContentHeader = Content
+legalNoticeSendContent = Mozilla receives an encrypted copy of the file you upload but we cannot access the content or name of your encrypted file.  Depending on how you use Send, files are stored for a maximum of either 24 hours or 7 days.  If you choose a download cap, the file may be deleted sooner.
+legalNoticeSendDataHeader = Data on your device
+legalNoticeSendData = So that you can check status or delete files, basic information about your uploaded files is stored on your local device.  This includes our identifier for the file, the filename, and the file’s download URL. This is cleared if you delete your uploaded file or upon visiting Send after the file expires.
+legalNoticeSendPersonalHeader = Personal data
+legalNoticeSendPersonal = The following is necessary to provide the service
+legalNoticeSendNonPersonalHeader = Non-personal data
+legalNoticeSendNonPersonal = We receive the following to improve our service and performance
+legalNoticeSendThirdPartyHeader = Third Party Services
+legalNoticeSendThirdParty = We use Google Cloud Platform
+legalNoticeSendIPHeader = IP addresses
+legalNoticeSendIP = We receive IP addresses of downloaders and uploaders as part of our standard server logs.  These are retained for 90 days, and for that period, may be connected to activity of a file’s download URL.  Although we develop our services in ways that minimize identification, you should know that it may be possible to correlate the IP address of a Send user to the IP address of other Mozilla services with accounts; and if there is a match, this could identify the account email address.
+legalNoticeSendFXAHeader = Firefox Account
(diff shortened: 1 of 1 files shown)


public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
 maxPasswordLength = Maximum password length: { $length }
 # A short status message shown when there was an error setting the password
 passwordSetError = This password could not be set
+#
```
flodolo: What's up will these commented lines?

If they're not needed, please remove them. This is going to show up as one comment for the next valid string.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+#  passwordSetError = This password could not be set
+#
+#
+#
+# vnext ->
+#
+#
+#
+introTitle = Simple, private file sharing
+introDescription = Firefox Send lets you share files with end-to-end encryption and a link that automatically expires. So you can keep what you share private and make sure your stuff doesn't stay online forever.
```
flodolo: Passing by observation: "your stuff"? :-\

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+notifyUploadEncryptDone = Your file is encrypted and ready to send
+archiveExpiryInfo = Expires after { $downloadCount } or { $timespan }
```
flodolo: Add a comment explaining what's replacing these placeables.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+    }
+timespanWeeks = { $num ->
+        [one] 1 week
+       *[other] { $num } weeks
+    }
+fileCount = { $num ->
+    [one] 1 file
+   *[other] { $num } files
+}
+totalSize = Total size: { $size }
```
flodolo: What's $size? How is it formatted?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+tooManyFiles = { $count ->
+    *[other] Only { $count } files can be uploaded at a time.
+}
+# count will always be > 10
+tooManyArchives = { $count ->
+    *[other] Only { $count } archives are allowed.
+}
+expiredTitle = This link has expired.
+# Firefox Send is a brand name and should not be localized.
+notSupportedDescription = Firefox Send will not work with this browser. Send works best with the latest version of Firefox, and will work with the current version of most browsers.
```
flodolo: Since we're at this, we should create a term with this note

```
-send-brand = Firefox Send
```

And use it all the new strings instead of `Firefox Send` (do not modify old strings though)

```
notSupportedDescription = { -send-brand } will not work with this browser. Send works best with the latest version of Firefox, and will work with the current version of most browsers.
```

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+downloadFirefox = Download Firefox
+legalTitle = Send Privacy Notice
```
flodolo: Same here, let's avoid confusion by introducing a term at the beginning of the file

```
-send-short-brand = Send
legalTitle = { -send-short-brand } Privacy Notice
```

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+legalDateStamp = Version 1.0, dated March 12, 2019
+legalNoticeSend = When Mozilla receives information from you, our <a>Mozilla Privacy Policy</a> describes how we handle that information. Below are the top things you should know about Firefox Send.  You can also view the code <a>here</a>.
```
flodolo: Get rid of double spaces after periods, they're not relevant, or have any effect.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+legalNoticeSendPersonalHeader = Personal data
+legalNoticeSendPersonal = The following is necessary to provide the service
+legalNoticeSendNonPersonalHeader = Non-personal data
+legalNoticeSendNonPersonal = We receive the following to improve our service and performance
+legalNoticeSendThirdPartyHeader = Third Party Services
+legalNoticeSendThirdParty = We use Google Cloud Platform
+legalNoticeSendIPHeader = IP addresses
+legalNoticeSendIP = We receive IP addresses of downloaders and uploaders as part of our standard server logs.  These are retained for 90 days, and for that period, may be connected to activity of a file’s download URL.  Although we develop our services in ways that minimize identification, you should know that it may be possible to correlate the IP address of a Send user to the IP address of other Mozilla services with accounts; and if there is a match, this could identify the account email address.
+legalNoticeSendFXAHeader = Firefox Account
```
flodolo: Make `Firefox Account` a term and reference it across the strings.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+}
+accountBenefitSync = Manage shared files from any device
+accountBenefitMoz = Learn about other Mozilla services
+signOut = Sign Out
+okButton = OK
+downloadingTitle = Downloading
+noStreamsWarning = This browser might not be able to decrypt a file this big.
+noStreamsOptionCopy = Copy the link to open in another browser
+noStreamsOptionFirefox = Try our favorite browser
+noStreamsOptionDownload = Continue with this browser
```
flodolo: Please add a blank line at the end.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+orClickWithSize = or click to to send up to { $size }
+addPassword = Protect with password
+emailPlaceholder = Enter your email
+signInSizeBump = Sign in to send up to { $size }
+signInButton = Sign in/up
+accountBenefitTitle = Create a Firefox Account or sign in
+accountBenefitLargeFiles = Share files up to { $size }
+accountBenefitDownloadCount = Share files with more people
+accountBenefitTimeLimit = { $count ->
+    *[other] Keep links active for up to { $count } days
```
flodolo: Even if it's not used by English, let's add the `one` form here.

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+legalNoticeSendInteractionHeader = Interaction data
+legalNoticeSendInteraction = This includes information such as number of people sending and receiving files, number of files uploaded and approximate file sizes, percentage of file downloaders who become uploaders, how people engage with the website (time spent, clicks, referrer information, site exit path, use of passwords)
+legalNoticeSendTechnicalHeader = Technical data
+legalNoticeSendTechnical = This includes information such as operating system, browser, language preference, country, timestamps, duration for file transfer, reasons for errors, reasons for file expiration.
+# A short representation of a countdown timer containing the number of days, hours, and minutes remaining as digits, example "2d 11h 56m"
+expiresDaysHoursMinutes = { $days }d { $hours }h { $minutes }m
+addFilesButton = Select files to upload
+uploadButton = Upload
+dragAndDropFiles = Drag and drop files
+orClickWithSize = or click to to send up to { $size }
```
flodolo: Why is this message split into two strings?

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,146 @@ passwordIsSet = Password set
...
+addPassword = Protect with password
+emailPlaceholder = Enter your email
+signInSizeBump = Sign in to send up to { $size }
+signInButton = Sign in/up
```
flodolo: Why is this using sentence case and the SignOut button title case?

flodolo:
Additional side note: we're encouraging developers to use kebab-case for string IDs. I guess it's too late for Send, but it's worth noting for future projects.

public/locales/en-US/send.ftl
johngruen: Strings come from content strategist

johngruen: this comes from content strategy. i'm not super thrilled with this one but it's been through like 10 rounds of edits

public/locales/en-US/send.ftl
johngruen: I think danny is trying to differentiate between old strings, strings that are shared across old and new as well as new strings. I will remove the annotations. Cool if i leave a space?

johngruen: Looks like i can add FTL section heads

https://github.com/l20n/l20n-syntax-fixtures/blob/master/parser/ftl/04-sections.ftl

public/locales/en-US/send.ftl
johngruen: good catch

public/locales/en-US/send.ftl
johngruen: Can you clarify?

public/locales/en-US/send.ftl
johngruen: I added a comment in my PR...this is total allowable size measured in terms of GB. this will return something like 2.5GB

public/locales/en-US/send.ftl
flodolo: Sorry, reference is here. That's a very old (and obsolete) project
https://projectfluent.org/fluent/guide/

public/locales/en-US/send.ftl
flodolo: Right, the question is more about how we format that number. Do we have formatting by locale, do we just assume $number$unit, no spaces?

public/locales/en-US/send.ftl
flodolo: This

```
accountBenefitTimeLimit = { $count ->
     [one] Keep links active for up to 1 day
    *[other] Keep links active for up to { $count } days
}

File: public/locales/en-US/send.ftl (hunk 1 above)
Comment (johngruen) [reply public/locales/en-US/send.ftl]: Per danny's note above, the commented lines are currently in prod. Since they are already localized, i think the GH diff does not matter here

File: public/locales/en-US/send.ftl (hunk 4 above)
Comment (johngruen) [reply public/locales/en-US/send.ftl]: https://github.com/mozilla/send/blob/master/app/utils.js#L98

Comment (flodolo) [reply public/locales/en-US/send.ftl]: OK, that's not great, but we can look into it later. This would need at least one string to define how the string is formatted (e.g. space or not, and which kind, between number and unit). Normally we expose for localization also the units.

**Comment by dannycoates** [comment]
@flodolo thanks. I think I've addressed all the comments. Can you have another look?

File: public/locales/en-US/send.ftl [hunk 13]
```diff
@@ -117,3 +117,87 @@ passwordIsSet = Password set
 maxPasswordLength = Maximum password length: { $length }
 # A short status message shown when there was an error setting the password
 passwordSetError = This password could not be set
+
+## New strings for the vNext version of Firefox Send
+
+# Firefox Send, Send and Firefox Account are proper names and should not be localized
+# If used in context, the words send and account should be localized
```
flodolo: Correction: "Firefox Account" is localized. I would just put

```
# Firefox Send, Send, Firefox, Mozilla are proper names and should not be localized
```

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,87 @@ passwordIsSet = Password set
...
+totalSize = Total size: { $size }
+# the next line after the colon contains a file name
+copyLinkDescription = Copy the link to share your file:
+copyLinkButton = Copy link
+downloadTitle = Download files
+downloadDescription = This file was shared via { -send-brand } with end-to-end encryption and a link that automatically expires.
+trySendDescription = Try { -send-brand } for simple, safe file sharing.
+# count will always be > 10
+tooManyFiles = { $count ->
+    *[other] Only { $count } files can be uploaded at a time.
```
flodolo: See comment I had for the other string. For clarity, I would still add the `[one]` case, even if it's not used in English (also in the next one).

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,87 @@ passwordIsSet = Password set
...
+legalDateStamp = Version 1.0, dated March 12, 2019
+# A short representation of a countdown timer containing the number of days, hours, and minutes remaining as digits, example "2d 11h 56m"
+expiresDaysHoursMinutes = { $days }d { $hours }h { $minutes }m
+addFilesButton = Select files to upload
+uploadButton = Upload
+# the first part of the string 'Drag and drop files or click to send up to 1GB'
+dragAndDropFiles = Drag and drop files
+# the second part of the string 'Drag and drop files or click to send up to 1GB'
+# size is a localized number followed by a unit of bytes, ex. 2.5GB
+orClickWithSize = or click to to send up to { $size }
```
flodolo: double 'to'

public/locales/en-US/send.ftl
```diff
@@ -117,3 +117,88 @@ passwordIsSet = Password set
 maxPasswordLength = Maximum password length: { $length }
 # A short status message shown when there was an error setting the password
 passwordSetError = This password could not be set
+
+## New strings for the vNext version of Firefox Send
+
+# Firefox Send, Send, Firefox, Mozilla are proper names and should not be localized
+# If used in context, the words send and account should be localized
```
flodolo: To be clear, please remove the second line of the comment

dannycoates:
Thanks @flodolo

flodolo:
I clearly shouldn't be doing review when I can't focus 100% on it. I never asked to remove the term for `Firefox Account`, exposing it as `{-firefox} Account` is wrong :-\

What I meant was to remove the comment, since *account* can be translated.

This is just a reminder for next time we add new strings.

PR #1190, opened 2019-03-05
fix issues based on feedback

johngruen:


public/locales/en-US/send.ftl
```diff
@@ -117,13 +117,10 @@ passwordIsSet = Password set
...
-#
-#
-#
-# shared ->
-#
-#
-#
+
+# Denotes strings shared across versions
+ [[ shared ]]
```
flodolo: This is not a supported syntax anymore. We don't have section names, we can have section comments (with two hashes).

```
 ## Strings shared across different versions of Send
```

public/locales/en-US/send.ftl
```diff
@@ -117,13 +117,10 @@ passwordIsSet = Password set
...
+
 #  title = Firefox Send
```
flodolo: There are still commented out strings. What the outcome you hope to get by doing this? Right now this will be considered a single huge comment for the next string `send-brand`, and displayed accordingly in Pontoon.

By commenting them, you'll still lose all translations. Either you keep them uncommented, or you remove them.

public/locales/en-US/send.ftl
```diff
@@ -168,16 +165,20 @@ passwordSetError = This password could not be set
...
-#
-#
-#
-# vnext ->
-#
-#
-#
+
+# New strings for the vNext version of Firefox Send
+[[ vnext ]]
```
flodolo: Same here, unsupported section name. This should just be

```
## New strings for the vNext version of Firefox Send
```

public/locales/en-US/send.ftl
```diff
@@ -195,13 +196,13 @@ fileCount = { $num ->
     [one] 1 file
    *[other] { $num } files
 }
+# $size refers to megabytes or gigabytes
```
flodolo: My question is more about: how do you format the number? Can you point me to the code?

I have the feeling it's old code, but it would be nice to double check. Different locales will have different formatting options, potentially names for units (it can also be a follow-up if needed, to speed things up)

public/locales/en-US/send.ftl
```diff
@@ -212,51 +213,33 @@ tooManyArchives = { $count ->
...
-legalNoticeSendInteractionHeader = Interaction data
-legalNoticeSendInteraction = This includes information such as number of people sending and receiving files, number of files uploaded and approximate file sizes, percentage of file downloaders who become uploaders, how people engage with the website (time spent, clicks, referrer information, site exit path, use of passwords)
-legalNoticeSendTechnicalHeader = Technical data
-legalNoticeSendTechnical = This includes information such as operating system, browser, language preference, country, timestamps, duration for file transfer, reasons for errors, reasons for file expiration.
 # A short representation of a countdown timer containing the number of days, hours, and minutes remaining as digits, example "2d 11h 56m"
 expiresDaysHoursMinutes = { $days }d { $hours }h { $minutes }m
 addFilesButton = Select files to upload
 uploadButton = Upload
+# The next two lines are two parts of a split title. You can see them at https://send2.dev.lcip.org
 dragAndDropFiles = Drag and drop files
```
flodolo: Localizers don't see the file, they see one string and comment at a time

```
# First part of a split title (see https://send2.dev.lcip.org)
dragAndDropFiles = Drag and drop files
# Second part of a split title (see https://send2.dev.lcip.org)
 orClickWithSize = or click to to send up to { $size }
```

public/locales/en-US/send.ftl
johngruen: so i talked to @dannycoates these are strings that don't need to be localized b/c they already are. however since they show up as new in the github diff we didn't want to cause confusion. I've spoken with danny and we'll merge my PR into his branch as is and remove the comments in his branch

public/locales/en-US/send.ftl
johngruen: Yep: https://github.com/mozilla/send/blob/master/app/utils.js#L98

johngruen: I'll carry this over as a comment on danny's PR so we can merge this branch into his

public/locales/en-US/send.ftl
johngruen: I'll move this comment over to danny's PR so we can merge this one and resolve there

flodolo: You should remove them completely, or leave them as they are (uncommented). Commenting them out is the only option that doesn't work.

flodolo: OK, I think I got confused by the comment. You plan to leave them as they are (remove the comments).

johngruen: yes

Issue #1192, opened 2019-03-06
[Send v2] The Privacy page is not opened when clicking the "Privacy" link 


SoftVision-CosminMuntean:
*[Affected Platforms]:*
- All Windows
- All Mac
- All Linux

*[Affected Versions]:*
- Firefox 65 and up

*[Steps to reproduce]:*
1. Open the Firefox browser and navigate to "https://send2.dev.lcip.org/" website.
2. Click the "Privacy" link from the footer of the page.
3. Observe the behaviour.

*[Expected result]:*
- You are successfully redirected to the Privacy page.

*[Actual result]:*
- The Privacy page is not opened even if the URL is changed into "https://send2.dev.lcip.org/legal".

*[Notes]:*
- If you wait ~5 minutes after clicking the "Privacy" link, you are redirected to the "504 Gateway Timeout Error" page.
- Attached a screen recording with the issue:
![privacy](https://user-images.githubusercontent.com/19226534/53871851-bc51c380-4005-11e9-8869-f5860b1b77ee.gif)

johngruen:
@SoftVision-CosminMuntean i believe this was caused by us leaving the legal strings out of localization...we're adding them as en-only in html for now while we wait on legal to localize them for us. This pr should resolve https://github.com/mozilla/send/pull/1201

dannycoates:
fixed

Softvision-AndreeaCupsa:
I have verified this issue and is no longer reproducible on the https://send2.dev.lcip.org on Windows 10 x64, Mac 10.14 and Arch Linux.

PR #1201, opened 2019-03-07
Html pn

johngruen:
@dannycoates this adds the legal text in as HTML. PR-ing into master since it looks like you're gearing up to merge vnext onto master anyway


johngruen:
Fixes #1192

PR #1242, opened 2019-03-13
[Integration-Test] Add test to ensure that file size stays consistent

### LINKED ISSUE #1130: [Integration-Test] [Send-v2] Test uploaded file size matches download
## Test overview
* This test should check that the downloaded file size matches the uploaded file size.

### Test Steps
Action | Result
------------ | -------------
 Navigate to Send Homepage | Send homepage loads
Upload a file and click the "Upload" button | File is Uploaded
Navigate to file download URL and download file | File downloads
Check file size on disk | File size matches uploaded file

The file size should be checked on disk for both the uploaded and downloaded file, not using the size displayed via the UI. This test should probably exist within the ```download-tests.js``` file.

mgalicer:
fixes #1130

### DIFF
--- a/test/integration/download-tests.js
+++ b/test/integration/download-tests.js
@@ -56,4 +56,29 @@ describe('Firefox Send', function() {
       'Expires after 1 download'
     );
   });
+
+  it('should ensure that the downloaded file size matches the uploaded file size', function() {
+    browser.chooseFile(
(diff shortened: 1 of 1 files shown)


dannycoates:
Thank you @mgalicer 😄

PR #1247, opened 2019-03-14
Proposed a few grammatical and typo fixes

mikeshatch:
Edits:

mimimum ==> minimum
anonmyized ==> anonymized 
listene… ==> listener
decryptin… ==> decrypting


dannycoates:
Thanks @mikeshatch. I would merge this but this file is auto-generated so the edits would get overwritten

mikeshatch:
You're welcome @dannycoates. I'm not familiar with this. Just out of curiosity, how do I tell beforehand if a repository's file is auto-generated?

PR #1248, opened 2019-03-14
Update CONTRIBUTORS

787954759:


dannycoates:
thanks @787954759, we've got a script that does this for us.

PR #1259, opened 2019-03-17
Updated file size limit (as of March 2019)

co16353sidak:
in FAQ.md,
The file size limit was previously marked as 2GB (As of April 2018), but now it supports 2.5GB (1GB for non signed in users)
So I have made an attempt to change this to be according to the current file size limits.

I did not create an issue first as I am hoping this fix is trivial and doesn't require an issue first.


dannycoates:
thank you @co16353sidak 😄

PR #1294, opened 2019-04-09
Rename circle.yml to .circleci/config.yml

drazisil:
Corrected config filename and path. This should get it building correctly on CircleCI

Hi folks. Stumbled across this project. This PR should help get it building on CircleCI.


dannycoates:
circle.yml works fine as is, but sure why not. Thanks @drazisil

drazisil:
@dannycoates It's not _supposed to_ with CircleCI 2.0 and is very likely to stop working once someone notices that. I wanted to proactively not have your builds break when that happens :)

PR #1299, opened 2019-04-19
Prevent possible download counter race condition

slootjes:
This very small change leverages the _hincrby_ function of Redis instead of setting the download count manually. Even though it would be very rare, it could lead to a race condition where multiple requests will set the same download count which causes it be counted as just 1 download.


dannycoates:
Thank you @slootjes!

PR #1320, opened 2019-05-01
added signin button color experiment

dannycoates:


fzzzy:
Looks good. I like the purple one.

Issue #423, opened 2017-08-04
Please Support QRCode.


ousui:
I want download the send file by my phone.
But the url is too long to input.
Can Support QRCode?

abdulhannanali:
@pdehaan This feature request seems good and useful. Any thoughts? I want to start working on this, If this feature is something mozilla people deem useful. I think it should be a hit with Chinese users.

pdehaan:
@abdulhannanali It definitely sounds interesting, and I know we've considered it for other projects (https://github.com/mozilla/fxa-features/issues/47). I'd probably hold off on implementation until we hear back from product management on if this is something we'd like to try in A/B testing or as a new feature.

pdehaan:
@vladikoff shane-tomlinson Did we ever implement QR code support in FxA in any way (intern project, etc)? I thought Sai was working on it at some point, but maybe I'm horribly mis-remembering.

ehuggett:
It seems it was at least considered for pairing devices? https://wiki.mozilla.org/Services/Sync1_5/FxASyncPairing#Pairing_mechanisms

Firefox mobile appears to be able to read QR codes https://support.mozilla.org/en-US/kb/scan-qr-codes-firefox-android https://support.mozilla.org/en-US/kb/scan-qr-codes-firefox-ios , although any QR code reader would of course be suitable on mobile long long as it can pass the url to a browser.

gvanwaelvelde:
Support for QR code sounds interesting, but since a file can be downloaded only once, I'm not sure it would be very useful?

ehuggett:
You would have to decide you wanted the file on your mobile while the download page is displayed on another device, before you click download (you would click QR code instead).

I can't see any obvious way of sending a QR code without re-creating the "how do I send files securely" problem, but the receiver of a link has all the information required to create them.

johngruen:
@ousui could you give a concrete example of when you would want a QR code? Is this sending to a friend or sending a file to yourself?

ousui:
@johngruen 
In China, many people use PC with vpn/proxy to download files like apk, videos. 
When I have not  usb-cable, I'll download these files with send.firefox.com. But the url is too long to input.

ehuggett:
I was about to suggest a firefox addon, but the first 3 i have looked at from the addons search results (including "JavaScript QR Code, Make QR Codes in pure JavaScript" and one marked "featured") rely on or expose the data to an external service (some of them over plain http!) and don't necessarily make this obvious to the user.

[...]

PR #1328, opened 2019-05-07
Qr code download link proposal

oliver-dvorski:
So in hopes to bump #423 a bit, I figured why not try to implement that feature myself.

Since I have no idea what I'm doing and since I've never heard of Choo before diving into this, I tried oversimplifying the whole thing and coupled a QR code view to the `copy to clipboard` action. This is what that looks like:

screencast-localhost-8080-2019 05 07-21-14-06

Any thoughts? I'm expecting to trigger *everyone* with the way I wrote this so don't hold back, just let it out

Sorry for the broken gif, it's late and I want to go home 😄


dannycoates:
Nice start @oliver-dvorski! We'll want to tweak the ui some, but this is great. I'll be able to give you more feedback next week.

oliver-dvorski:
@dannycoates Awesome 😊
Can't wait to refine this

cmpadden:
Hi @oliver-dvorski, while I don't have any feedback on this pull request I wanted to say thank you for your proposal.

I often find myself wishing that the Firefox Send URLs were either shorter or had an easy way to copy them to other devices -- this would be great! Thanks!

oliver-dvorski:
@cmpadden Thanks! That means a lot. I too can't wait to get this rolling so I can finally kill [makeazip.com](https://makeazip.com) 😄 and let the big boys provide a much better service than I ever could

app/routes.js
```diff
@@ -10,6 +10,7 @@ module.exports = function(app = choo()) {
   app.route('/legal', body(require('./ui/legal')));
   app.route('/error', body(require('./ui/error')));
   app.route('/blank', body(require('./ui/blank')));
+  // app.route('/qr', body(require('./ui/displayQR')))
```
akshaynexus: Better to remove this commented out route if its not in use

oliver-dvorski: Good point! Fixed that now

cjbarth:
This is also super useful for sending stuff from Android to iOS, like photos. What else is needed to get this landed?

fzzzy:
The only thing required for this to land is for someone to have time to review it. There's nobody full time on send right now. I might be able to take a look at it next week.

cjbarth:
@oliver-dvorski It would be nice if there were no merge conflicts when this does get reviewed, hopefully next week. Would you be in a position to fix this merge conflict?

oliver-dvorski:
Should be doable. Tomorrow I'll be heading back from vacation so I should
have that fixed soon

Chris Barth <notifications@github.com> schrieb am Mo., 7. Okt. 2019, 19:51:

oliver-dvorski:
Alrighty, I think I got myself out of that git mess I made. Should be squeaky clean now

cjbarth:
It looks like @dannycoates updated dependencies in this project and messed up the lock file again :( Maybe just leave the lock file out of this PR and let the maintainers add the lock for the dependencies that you need when they do their regular package updates?

oliver-dvorski:
@cjbarth Brilliant Idea! Unfortunately, the test pipeline is checking to see if the two files are in sync and when they're not - it throws an error. So I thought I'll just update the thing instead and just push whenever. Which is still cool, because now that I rebased on master, I see that there's a test that's failing so I can address that

PR #1341, opened 2019-05-28
added master logo and trailhead strings

dannycoates:


public/locales/en-US/send.ftl
```diff
@@ -143,3 +143,5 @@ shareLinkDescription = Share the link to your file:
 shareLinkButton = Share link
 # $name is the name of the file
 shareMessage = Download “{ $name }” with { -send-brand }: simple, safe file sharing
+trailheadPromo = There is a way to protect your privacy. Join Firefox.
+learnMore = Learn more
```
clouserw: The approved string ends with a period.  (I agree it looks weird)

flodolo:
Actually, I don't think I can merge in this repo :-\

PR #1362, opened 2019-07-05
Check if browser is not Firefox

### LINKED ISSUE #1360: Can we limit the display of the promo banner to only non-firefox userAgents?
It's just a thought. Personally, I'm super annoyed by the thing but that isn't really relevant. I'd just like to know if this is a design decision or not.

Sorry if I'm posting in the wrong place, I don't see a forum anywhere so... here we are

wadie:
closes #1360

### DIFF
--- a/app/ui/body.js
+++ b/app/ui/body.js
@@ -8,7 +8,7 @@ function banner(state) {
     return; // server side
   }
   const show =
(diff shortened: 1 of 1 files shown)


clouserw:
Thanks for the patch.  I mentioned in #1360 that this was done by design though, so I'm going to close this.

PR #1365, opened 2019-07-11
Deploy to Heroku button added

yogishb608:


clouserw:
Thanks for the patch, but this isn't something we want to land (and maintain).  Closing this.

Issue #1377, opened 2019-07-31
Investigate failed uploads / downloads


dannycoates:
There's been some mention of hanging uploads and downloads on the web. This for example: https://old.reddit.com/r/firefox/comments/cjpmhm/real_world_review_for_firefox_send/

I may have reproduced it and got this error in the browser:
<img width="561" alt="Screen Shot 2019-07-31 at 9 26 56 AM" src="https://user-images.githubusercontent.com/87619/62229886-8b3bac00-b375-11e9-92be-fc9c7dfb9063.png">

Note: Send currently isn't staffed. Add a 👍 if this issue affects you.

LoneFenris:
#1280 potentially related on the download side.

aakay:
Hi @dannycoates , thanks for the investigation and the ticket.

I'm the author of the post. If there's something I can help, please don't hesitate to contact me.

dannycoates:
@aakay thanks for your patience. Our next release will hopefully fix most upload issues and provide us better error reporting to help identify any others. If you're interested I welcome you to try it out before release on our testing site https://send2.dev.lcip.org

aakay:
Hey @dennycoates, thanks a lot for the fix. I’d love to help whenever I can. Sure, I’ll try it when I need to send a file, as I’ve done till now :).

dannycoates:
v3.0.16 was released yesterday. Early data suggests our internal connection issue is fixed but "regular" connection issues between the device and us can still be improved with a more robust retry mechanism. I'm closing this in favor of #1380

PR #1378, opened 2019-08-07
Use resumable uploads to GCS

dannycoates:
The non-resumable GCS upload uses a route with a 30 minute timeout. The resumable route doesn't have this timeout. This should address part of #1377

This PR also improves the UI when the websocket closes early for any reason. Instead of hanging it will go to the error page.

`@google-cloud/storage` module implements resumable uploads in a way that requires fs write access to a config file. We should watch this to make sure it's behaving ok, otherwise we may need to modify it to not use a file.


server/bin/dev.js
```diff
@@ -3,7 +3,7 @@ const routes = require('../routes');
 const pages = require('../routes/pages');
 const tests = require('../../test/frontend/routes');
 const express = require('express');
-const expressWs = require('express-ws');
+const expressWs = require('@dannycoates/express-ws');
```
fzzzy: Are there changes in your copy of express-ws that are needed? Or was this just for debugging?

dannycoates: I updated the `ws` dependency to the latest major version. We've been using a fork of express-ws since the beginning, I just finally pushed it to npm

Issue #1354, opened 2019-06-15
Wrong direct link when share in facebook


tranlehaiquan:
I have a send link with password. When I shared it through Facebook messenger it will routing to another link not which i want.

My file: https://send.firefox.com/download/fb01d0252e05f07b/#IT3oR8BAx5nkI5IvrweCew
Facebook will direct to: https://send.firefox.com/download/fb01d0252e05f07b/?fbclid=IwAR3UM1S96aw8_8WQfTgHdqkQi9nmOFrua9xz6RZB0Jdmcea6yjKTE9O7v54#IT3oR8BAx5nkI5IvrweCew

geekykant:
I would agree more to it for Facebook to solve this query string problem. The unique **tracking parameter** added by Facebook, just fails itself to redirect properly.

tranlehaiquan:
But i don't know if it is right or wrong. It redirect me to same file with password. I lost 30 minutes to figure that out : )).

clouserw:
It looks like this is something facebook adds for tracking.  It does break the download link though so it's a legitimate bug -- we should strip out or ignore any extra URL params

PR #1408, opened 2019-10-15
Strip tracking URL params

drakkein:
This PR strips all URL parameters between file id and secret key before password validation.

Connected issue https://github.com/mozilla/send/issues/1354


drakkein:
The `should show progress when a file is uploading` integration test seems to not pass on CI even if it's green locally. Can any of maintainers check this?

PR #1422, opened 2019-12-04
Implement password requirements

sesquipedalian-dev:
Make (configurably) passwords mandatory for uploads and implement password requirements checking. 

- add values to config to make passwords mandatory, and to provide a list of password validation rules by name
- Updated download on server to abandon early (400) when PASSWORD_REQUIRED is true if the file was somehow saved without a password.  This should backstop attempts to modify the frontend code to not assign a password, since the file will ultimately not be downloadable. 
- Update frontend to make the password field always appear if PASSWORD_REQUIRED is true
- updated frontend so that upload button changes color and doesn't do anything, if a password is required and the password doesn't meet the parameters
- Fixed password message to allow multiline text, for multiple error cases 
- added more password validation options, such as by character classes and minimum length


sesquipedalian-dev:
One current bug: if you don't enter a password after getting to that screen, it will let you upload with a blank.  The validation only happens on change of the password field.  The backend will also let you download with a password of `''`.

sesquipedalian-dev:
whoops, didn't mean to send this to the upstream just yet ^^;

PR #1431, opened 2020-01-02
Add new locales (an, gn, mix)

flodolo:
All three are at 100%, quickly tested them on the DEV server.


dannycoates:
One unrelated issue to fix around geoip then I'll do a release

PR #1455, opened 2020-03-24
update promo to be for the vpn

johngruen:
Hey @dannycoates I didn't use l10n here because the VPN is US only at the moment. This seemed easy. LMK if you want me to make any changes.

### DIFF
--- a/app/ui/body.js
+++ b/app/ui/body.js
@@ -8,7 +8,9 @@ function banner(state) {
     return; // server side
   }
   const show =
-    !state.capabilities.standalone && !state.route.startsWith('/unsupported/');
+    !state.capabilities.standalone &&
+    !state.route.startsWith('/unsupported/') &&
+    state.locale === 'en-US';
   if (show) {
     return state.cache(Promo, 'promo').render();
   }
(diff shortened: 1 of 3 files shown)


app/ui/promo.js
```diff
@@ -24,11 +24,11 @@ class Promo extends Component {
             alt="Firefox"
           />
           <span class="ml-2 sm:ml-4 text-xs sm:text-base">
-            ${this.state.translate('trailheadPromo')}${' '}
+            ${`Like Firefox Send? Youʼll love our new VPN.`}
```
chrismore: @johngruen what about "Like Firefox Send? You'll love our new full-device VPN." -- just thought is is important to make sure they know this is a full-device product as anything with "Firefox" immediately gets associated in the consumer mind with browser-only or firefox-only.

johngruen: 👍

app/ui/body.js
```diff
@@ -4,7 +4,7 @@ const Header = require('./header');
 const Footer = require('./footer');
 
 function banner(state) {
-  if (state.layout) {
+  if (state.layout || state.locale !== 'en-US') {
```
dannycoates: can you move this addition down to the `show =` expression?

PR #1459, opened 2020-04-12
About Mozilla Send

bmoser05:
About send.firefox.com.


dannycoates:
Thanks @bmoser05 but I think our readme.md is a good enough intro :)

Issue #1478, opened 2020-06-10
Add a more clear disclaimer for users who may be downloading harmful content


claudijd:
Below is an example of what a user sees when they download a file.  However, I think we should be more clear about the fact that this file is not "from Mozilla" and that there should be some sort of guidance/warning of some sort to educate users about malicious/harmful content that Mozilla/Firefox do not control as well as better guidance, tooling, and process for abuse claims.

<img width="1064" alt="Screen Shot 2020-06-10 at 10 15 39 AM" src="https://user-images.githubusercontent.com/794283/84278920-5787f180-ab03-11ea-92e1-5a0b41d2a96a.png">

fzzzy:
I absolutely agree we should do this.

claudijd:
Note that part of this is interrelated to #1479

dannycoates:
started here 9891d1f0baf6ec459b2706bce81b56ba46e7bebc

PR #1488, opened 2020-07-02
An idea for a warning placement to remind users to think about who sent the link

claudijd:
Was playing around with Send code, thought I'd whip up a warning and see if it resonates.  I realized it's not localize, but wanted to share.

Screen Shot 2020-07-02 at 2 50 22 PM


claudijd:
This is somewhat in the spirit of #1478

fzzzy:
Yes, something like this is sorely needed, but @johngruen should rework the styles first.

PR #1496, opened 2020-07-24
Begin implementing a reporting mechanism

dannycoates:


public/locales/en-US/send.ftl
```diff
@@ -145,3 +145,33 @@ shareLinkButton = Share link
...
+# your language, you can add {$count} to your translations and use the
+# standard CLDR forms, or only use the form for [other] if both strings should
+# be identical.
+reportFile =
+    { $count ->
+        [one] Report this file as suspicious
+       *[other] Report these files as suspicious
+    }
+reportDescription = Help us understand what’s going on. What do you think is wrong with these files?
+reportUnknownDescription = Please go to the url of the link you wish to report and click “{ reportFile }”.
```
flodolo, line 170: This should display the plural form, since it's the default. Can you confirm that works?

public/locales/en-US/send.ftl
```diff
@@ -145,3 +145,33 @@ shareLinkButton = Share link
...
+reportButton = Report
+reportReasonMalware = These files contain malware or are part of a phishing attack.
+reportReasonPii = These files contain personally identifiable information about me.
+reportReasonAbuse = These files contain illegal or abusive content.
+reportReasonCopyright = To report copyright or trademark infringement, use the process described at <a>this page</a>.
+reportedTitle = File Reported
+reportedDescription = Thank you. We have received your report on this file.
```
flodolo, line 177: Should these two be "Files Reported" and "…these files" at this point?

public/locales/en-US/send.ftl
dannycoates: yes, it does

dannycoates:
Thanks @flodolo

app/controller.js
```diff
@@ -306,6 +307,26 @@ export default function(state, emitter) {
...
+      if (!file) {
+        // TODO
+        emitter.emit('pushState', '/error');
+        return render();
+      }
+      await reportLink(file.id, file.secretKey, reason);
+      render();
+    } catch (err) {
+      console.error(err);
+      if (err.message === '404') {
```
fzzzy, line 322: Does this actually happen every time?

dannycoates: No, 

`/api/report` returns

* 200 - when the link was already flagged or was successfully flagged. Note that the file may already be "dead"
* 400 - when the submitted `key` can't decrypt the metadata. This is to prevent reporting ids without knowing the full url
* 404 - when the id has fully expired and deleted from redis. An api delete just marks the id as dead now so that we can accept reports for files that have already hit max download count

If it's a 404 we don't want to say there was an error

